PR #1, opened 2018-06-19
Go implementation of SingularityNET Daemon

erickearns:
- New gRPC daemon supports proxying
 - gRPC+proto or gRPC+JSON to gRPC service
 - gRPC+JSON to JSON-RPC service
 - gRPC+JSON to process fork service
 - JSON-RPC to JSON-RPC service (for backward compatibility)
- Exposes HTTP 1.1 grpc-web interface
- Exposes /encoding HTTP route for clients to determine whether to use JSON or proto encoding
- Validates job invocations using snet-job-address/snet-job-signature HTTP headers
- Maintains configuration options from python daemon


tiero:
**tACK**

README.md
```diff
@@ -1 +1,61 @@
...
+
+These instructions are intended to facilitate the development and testing of SingularityNET Daemon. Users interested in
+deploying SingularityNET services using SingularityNET Daemon should install the appropriate binary as
+[released](#release).
+
+### Prerequisites
+
+* [Go 1.10+](https://golang.org/dl/)
+* [Dep 0.4.1+](https://github.com/golang/dep#installation)
+* [Node 8+ w/npm](https://nodejs.org/en/download/)
```
sectioneight: Should note here that Node 10 is *not* supported due to package issues with sha3

blockchain/blockchain.go
```diff
@@ -0,0 +1,182 @@
+//go:generate abigen --abi ../resources/Agent.abi --pkg blockchain --type Agent --out agent.go
+
+package blockchain
+
+import (
+	"bytes"
+	"crypto/ecdsa"
+	"encoding/hex"
+	"encoding/json"
+	"github.com/coreos/bbolt"
```
sectioneight: These aren't in standard gofmt/goimports stanzas, where 3rd party is separated from 1st party is separated from project-local imports. Suggest we standardize on the community standard.

blockchain/blockchain.go
```diff
@@ -0,0 +1,182 @@
...
+	"github.com/ethereum/go-ethereum/ethclient"
+	"github.com/ethereum/go-ethereum/rpc"
+	"github.com/singnet/snet-daemon/config"
+	"github.com/singnet/snet-daemon/db"
+	log "github.com/sirupsen/logrus"
+	"google.golang.org/grpc"
+)
+
+const (
+	JobPendingState    = "PENDING"
```
sectioneight: golint/govet should be complaining about exported fields not having docstring comments. we should follow community best practices and also run go vet / go lint as part of checkin.

blockchain/blockchain.go
```diff
@@ -0,0 +1,182 @@
...
+				log.WithError(err).Panic("error deriving private key")
+			} else {
+				privateKey = privKey
+				address = crypto.PubkeyToAddress(privateKey.PublicKey).Hex()
+			}
+		}
+
+		// Start event and job completion routines
+		jobCompletionQueue = make(chan *jobInfo, 1000)
+		go processJobCompletions()
```
sectioneight: Doing stuff like this in `init()` is frowned upon. We should avoid spawning goroutines or really doing much of anything as a side effect of importing a package. Let's talk offline about this pattern.

blockchain/blockchain.go
```diff
@@ -0,0 +1,182 @@
...
+		go processEvents()
+		go submitOldJobsForCompletion()
+	}
+}
+
+func GetGrpcStreamInterceptor() grpc.StreamServerInterceptor {
+	if config.GetBool(config.BlockchainEnabledKey) {
+		return jobValidationInterceptor
+	} else {
+		return noOpInterceptor
```
sectioneight: golint / govet should complain about "unnecessary else" here

blockchain/blockchain.go
```diff
@@ -0,0 +1,182 @@
...
+	}
+
+	log.Debug("unable to validate job invocation locally; falling back to on-chain validation")
+
+	// Fall back to on-chain validation
+	if validated, err := agent.ValidateJobInvocation(&bind.CallOpts{
+		Pending: true,
+		From:    common.HexToAddress(address)}, common.BytesToAddress(jobAddressBytes), v, r, s); err != nil {
+		log.WithError(err).Error("error validating job on chain")
+		return false
```
sectioneight: this "log error and return false" makes me think that we should be returning an `error` or `nil` so that the caller can tell what happened (and log it itself; a validity check shouldn't really be logging this much)

blockchain/blockchain.go
```diff
@@ -0,0 +1,182 @@
...
+		return nil
+	})
+
+	// If job is marked completed locally, reject
+	if job.Completed {
+		log.Debug("job already marked completed locally")
+		return false
+	}
+
+	v, r, s, err := parseSignature(jobSignatureBytes)
```
sectioneight: Four is probably too many things to be returning from a function. Suggest making it a struct with fields/methods on it. IOW, `(typ, error)` should be the return signature.

erickearns:
Replaced by #2

PR #6, opened 2018-07-19
Support for new signature scheme and new contracts packaging

erickearns:
Depends on https://github.com/singnet/alpha-blockchain/pull/26

This isn't a good solution for longterm; currently hardcode the checksum of the old agent's bytecode for comparison.


blockchain/blockchain.go
```diff
@@ -1,12 +1,16 @@
...
 
 package blockchain
 
 import (
 	"bytes"
+	"context"
 	"crypto/ecdsa"
 	"encoding/hex"
 	"encoding/json"
+
```
sectioneight: did you manually do this, or is your goland still doing something weird? should be in the first stanza, not separated (and should happen automatically)

blockchain/blockchain.go
```diff
@@ -50,13 +55,33 @@ func init() {
...
-		if a, err := NewAgent(common.HexToAddress(config.GetString(config.AgentContractAddressKey)), ethClient); err != nil {
+		if a, err := NewAgent(agentAddress, ethClient); err != nil {
 			log.WithError(err).Panic("error instantiating agent")
 		} else {
 			agent = a
 		}
 
+		// Determine "version" of agent contract and set local signature hash creator
+		if bytecode, err := ethClient.CodeAt(context.Background(), agentAddress, nil); err != nil {
+			log.WithError(err).Panic("error retrieving agent bytecode")
```
sectioneight: We should (in a future change) talk about removing these panics

blockchain/blockchain.go
```diff
@@ -50,13 +55,33 @@ func init() {
...
+		} else {
+			bcSum := md5.Sum(bytecode)
+			if bytes.Equal(bcSum[:], []byte{244, 176, 168, 6, 74, 56, 171, 175, 38, 48, 245, 246, 189, 0, 67, 200}) {
+				sigHasher = func(i []byte) []byte {
+					return crypto.Keccak256([]byte{0x19}, []byte("Ethereum Signed Message:\n32"),
```
sectioneight: Suggest making these `[]byte("Ethereum ...)` things top-level vars/const (they can't be const as-is because they're slices) to clarify that these don't change based on input

blockchain/blockchain.go
```diff
@@ -50,13 +55,33 @@ func init() {
...
+						crypto.Keccak256(i))
+				}
+			} else {
+				sigHasher = func(i []byte) []byte {
+					return crypto.Keccak256([]byte{0x19}, []byte("Ethereum Signed Message:\n420x"),
```
sectioneight: Would be nice to explain what the magic `0x19` byte does here / why it's necessary

blockchain/blockchain.go
erickearns: this was the result of running goimports, but when i move it up below crypto/ecdsa and rerun, it stays there... not sure what's going on

blockchain/blockchain.go
```diff
@@ -50,13 +58,31 @@ func init() {
...
 		} else {
 			agent = a
 		}
 
+		// Determine "version" of agent contract and set local signature hash creator
+		if bytecode, err := ethClient.CodeAt(context.Background(), agentAddress, nil); err != nil {
+			log.WithError(err).Panic("error retrieving agent bytecode")
+		} else {
+			bcSum := md5.Sum(bytecode)
+			if bytes.Equal(bcSum[:], []byte{244, 176, 168, 6, 74, 56, 171, 175, 38, 48, 245, 246, 189, 0, 67, 200}) {
```
lcdupree: add comment

PR #8, opened 2018-07-20
Cobra-ify daemon command and remove panics

sectioneight:
Bubble up initialization/configuration errors instead of dumping
stacktraces. Break out daemon into init/start/stop functions.

Add pkg/errors to capture error context.


sectioneight:
Closing, will re-open from personal fork

PR #10, opened 2018-07-20
Add basic command-line daemon flags

sectioneight:
These are bound with cobra/viper, so you can still set them (and fetch
them) via viper (hence no code changes to config fetching). But now you
can also specify them on the command line, and `snetd --help` will show
available flags.

```
go run ./snetd/main.go --help
Usage:
  snetd [flags]

Flags:
  -b, --blockchain                 enable blockchain processing (default true)
  -c, --config string              config file (default "snetd.config")
      --db-path string             database file path (default "snetd.db")
      --ethereum-endpoint string   ethereum JSON-RPC endpoint (default "http://127.0.0.1:8545")
  -h, --help                       help for snetd
      --mnemonic string            HD wallet mnemonic
      --passthrough                passthrough mode
      --poll-sleep string          blockchain poll sleep time (default "5s")
  -p, --port int                   daemon listen port (default 5000)
      --service-type string        service type: one of 'grpc','jsonrpc','process' (default "grpc")
  -t, --type string                daemon type: one of 'grpc','http' (default "grpc")
      --wallet-index int           HD wallet index
      --wire-encoding string       message encoding: one of 'proto','json' (default "proto")
```

```
go run ./snetd/main.go -t foo
DEBU[0000] error reading config                          error="Config File \"snetd.config\" Not Found in \"[/home/aiden/src/go/src/github.com/singnet/snet-daemon]\""
ERRO[0000] Unable to initialize daemon                   error="unrecognized DAEMON_TYPE 'foo'"
exit status 2
```

### DIFF
--- a/snetd/cmd/flags.go
+++ b/snetd/cmd/flags.go
@@ -1,0 +1,48 @@
+package cmd
+
+import (
+	"github.com/singnet/snet-daemon/config"
+	log "github.com/sirupsen/logrus"
+	"github.com/spf13/cobra"
+)
+
+var (
+	cfgFile            = ServeCmd.PersistentFlags().StringP("config", "c", "snetd.config", "config file")
+	daemonType         = ServeCmd.PersistentFlags().StringP("type", "t", "grpc", "daemon type: one of 'grpc','http'")
+	blockchainEnabled  = ServeCmd.PersistentFlags().BoolP("blockchain", "b", true, "enable blockchain processing")
+	listenPort         = ServeCmd.PersistentFlags().IntP("port", "p", 5000, "daemon listen port")
+	ethEndpoint        = ServeCmd.PersistentFlags().String("ethereum-endpoint", "http://127.0.0.1:8545", "ethereum JSON-RPC endpoint")
+	mnemonic           = ServeCmd.PersistentFlags().String("mnemonic", "", "HD wallet mnemonic")
+	hdwIndex           = ServeCmd.PersistentFlags().Int("wallet-index", 0, "HD wallet index")
+	dbPath             = ServeCmd.PersistentFlags().String("db-path", "snetd.db", "database file path")
+	passthroughEnabled = ServeCmd.PersistentFlags().Bool("passthrough", false, "passthrough mode")
+	serviceType        = ServeCmd.PersistentFlags().String("service-type", "grpc", "service type: one of 'grpc','jsonrpc','process'")
+	wireEncoding       = ServeCmd.PersistentFlags().String("wire-encoding", "proto", "message encoding: one of 'proto','json'")
+	pollSleep          = ServeCmd.PersistentFlags().String("poll-sleep", "5s", "blockchain poll sleep time")
+)
+
+func init() {
+	rf := ServeCmd.PersistentFlags()
+	vip := config.Vip()
+
+	vip.BindPFlag(config.ConfigPathKey, rf.Lookup("config"))
+	vip.BindPFlag(config.DaemonTypeKey, rf.Lookup("type"))
+	vip.BindPFlag(config.BlockchainEnabledKey, rf.Lookup("blockchain"))
+	vip.BindPFlag(config.DaemonListeningPortKey, rf.Lookup("port"))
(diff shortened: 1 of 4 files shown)


snetd/cmd/flags.go
```diff
@@ -0,0 +1,48 @@
...
+	hdwIndex           = ServeCmd.PersistentFlags().Int("wallet-index", 0, "HD wallet index")
+	dbPath             = ServeCmd.PersistentFlags().String("db-path", "snetd.db", "database file path")
+	passthroughEnabled = ServeCmd.PersistentFlags().Bool("passthrough", false, "passthrough mode")
+	serviceType        = ServeCmd.PersistentFlags().String("service-type", "grpc", "service type: one of 'grpc','jsonrpc','process'")
+	wireEncoding       = ServeCmd.PersistentFlags().String("wire-encoding", "proto", "message encoding: one of 'proto','json'")
+	pollSleep          = ServeCmd.PersistentFlags().String("poll-sleep", "5s", "blockchain poll sleep time")
+)
+
+func init() {
+	rf := ServeCmd.PersistentFlags()
```
erickearns: What does `rf` stand for?

snetd/cmd/flags.go
```diff
@@ -0,0 +1,48 @@
...
+	vip := config.Vip()
+
+	vip.BindPFlag(config.ConfigPathKey, rf.Lookup("config"))
```
erickearns: Maybe we should rethink this config key name & associated variable names (currently referred to as "path" and "file" in various places). In reality, this populates with vip expects to be the filename prefix (without `.json`). There is also a path to search in, which is currently hardcoded as cwd.

snetd/cmd/flags.go
```diff
@@ -0,0 +1,48 @@
...
+	vip.BindPFlag(config.HdwalletMnemonicKey, rf.Lookup("mnemonic"))
+	vip.BindPFlag(config.HdwalletIndexKey, rf.Lookup("wallet-index"))
+	vip.BindPFlag(config.DbPathKey, rf.Lookup("db-path"))
+	vip.BindPFlag(config.PassthroughEnabledKey, rf.Lookup("passthrough"))
+	vip.BindPFlag(config.ServiceTypeKey, rf.Lookup("service-type"))
+	vip.BindPFlag(config.WireEncodingKey, rf.Lookup("wire-encoding"))
+	vip.BindPFlag(config.PollSleepSecsKey, rf.Lookup("poll-sleep"))
+
+	cobra.OnInitialize(func() {
+		vip.SetConfigName(*cfgFile)
```
erickearns: Does this need to be here and in config.go?

snetd/cmd/flags.go
sectioneight: "root flags". i can expand, but was trying to make the following block a tad more readable. naming suggestions?

snetd/cmd/flags.go
sectioneight: oh, good point, this isn't really a "path" at all, it's a filename. we should definitely change these flags.

would you like to see that in this PR, or a separate one?

snetd/cmd/flags.go
sectioneight: it functions correctly as-is, in that the `config/` package can be used without any command-line (or cobra) code and it will read from the location specified by the environment variable.

if the intention is that config is only used by the serving code, we can remove the call to "setconfigname" in config.go and only set it once here.

erickearns: My vote is remove it from config.go -- I don't think we need to maintain functionality in anything beyond snetd command currently

If someone were to use config.go outside the context of cobra, doesn't this PR remove all defaults that were previously set within viper (but are now set within cobra?)

snetd/cmd/flags.go
erickearns: I'd say this one for less PR noise & nothing gained from pushing intermediate code separately

snetd/cmd/flags.go
erickearns: I'm fine with it, just was curious (it's pretty clear from context what this is)

snetd/cmd/flags.go
sectioneight: Done

PR #12, opened 2018-07-20
Add information-only golint to test runs

sectioneight:
And start fixing some golints


sectioneight:
It doesn't address all of them (there are about 30 left), I decided I'd fix those in a future diff to avoid massively conflicting with the other PRs I have open. Should be able to do it soon though (and then we can turn them from warnings into errors if we want).

PR #15, opened 2018-07-23
Add SSL/TLS and LetsEncrypt autocert support

sectioneight:
This change adds automatic SSL support via LetsEncrypt ("http-01"
challenge, since tls-sni- challenges were disabled recently for
security issues).

If requested, the daemon now fetches certs and serves gRPC/HTTP over TLS
when requested.

### DIFF
--- a/snetd/cmd/serve.go
+++ b/snetd/cmd/serve.go
@@ -1,6 +1,7 @@
 package cmd
 
 import (
+	"crypto/tls"
 	"fmt"
 	"net"
 	"net/http"
@@ -19,6 +20,8 @@ import (
 	log "github.com/sirupsen/logrus"
 	"github.com/soheilhy/cmux"
 	"github.com/spf13/cobra"
+	"golang.org/x/crypto/acme/autocert"
+	"golang.org/x/net/http2"
 	"google.golang.org/grpc"
 )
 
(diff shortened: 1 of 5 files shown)


snetd/cmd/serve.go
```diff
@@ -81,34 +96,71 @@ func newDaemon() (daemon, error) {
...
+		if tlsConfig != nil {
+			log.Debug("enabling SSL support")
+			tlsc := tlsConfig.Clone()
+			// See: https://gist.github.com/soheilhy/bb272c000f1987f17063
+			tlsc.NextProtos = []string{http2.NextProtoTLS, "h2-14", "http/1.1"}
+
+			lis = tls.NewListener(lis, tlsc)
+		}
+		mux := cmux.New(lis)
+		grpcL := mux.Match(cmux.HTTP2HeaderField("content-type", "application/grpc"))
```
erickearns: We have this on prefix matching as content-type may be one of `application/grpc`, `application/grpc+proto`, `application/grpc+json` for our grpc clients

And we use MatchWithWriters/*SendSettings I think just for compatibility with Java clients... https://github.com/soheilhy/cmux#limitations 

I should have commented on both of these decisions specifically

sectioneight: Makes sense. I changed a bunch of this stuff while testing to try different permutations of protocols and didn't change it back to how it was originally.

Updated to revert changes and add comment as to why we're doing the prefix/settings stuff.

PR #19, opened 2018-07-24
Handle CORS

erickearns:


snetd/cmd/serve.go
```diff
@@ -174,13 +175,16 @@ func (d daemon) start() {
 		})
 
 		log.Debug("starting daemon")
+
 		go d.grpcServer.Serve(grpcL)
-		go http.Serve(httpL, httpHandler)
+		go http.Serve(httpL, handlers.CORS(
+			handlers.AllowedHeaders([]string{"Content-Type", "Snet-Job-Address", "Snet-Job-Signature"}))(httpHandler))
```
sectioneight: Would be nice to extract these headers into a var and that var here and below instead of copy-paste, so we don't forget to update.

sectioneight:
Discussed offline about package-lock.json -- not necessarily related to this change but needed for Eric to do work and not worth separating out to a separate PR

PR #23, opened 2018-07-25
Clarify control flow on gRPC interceptor

sectioneight:
This flows more logically and makes clear when errors are returned, when
the handler is run, and when the "CompleteJob" hook is done.

Generally, when a success state is reached, it's preferable to "return
nil" to indicate that no error is being returned, instead of "return
err" where "err" may be nil.

### DIFF
--- a/blockchain/interceptors.go
+++ b/blockchain/interceptors.go
@@ -30,15 +30,16 @@ func (p Processor) jobValidationInterceptor(srv interface{}, ss grpc.ServerStrea
 
 	jobSignatureBytes := common.FromHex(jobSignatureMd[0])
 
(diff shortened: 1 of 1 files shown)


erickearns:
This is 100% subjective. I'd prefer keeping refactors of existing code limited to functional changes; else different team members could go back and forth ad infinitum updating style to their liking. If it makes sense to make things more clear/readable/consistent in the context of actual logic changes, that's totally reasonable IMO.

sectioneight:
Ya, I wanted to talk about this, since I don't think it's 100% subjective. The current block looks a bunch of these:

```
if ! precondition
  reject request (with reason)
```

And this change makes it so:

```
if ! jobValid
  reject

if ! handled
  return err

success
```

Basically, preferring an "early return" and being consistent.

PR #37, opened 2018-09-21
Fix read default config if file exist

vsbogd:
Fix case when user doesn't specify configuration file but default configuration file ```snetd.config.json``` exists in filesystem. Before fix snet-daemon ignored this file.

### DIFF
--- a/snetd/cmd/flags.go
+++ b/snetd/cmd/flags.go
@@ -69,6 +69,7 @@ func init() {
 				fmt.Println("Error reading config:", *cfgFile, err)
 				os.Exit(1)
 			}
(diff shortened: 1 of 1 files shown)


snetd/cmd/flags.go
```diff
@@ -77,3 +78,11 @@ func init() {
 		log.Info("Cobra initialized")
 	})
 }
+
+func isFileExist(file string) bool {
+	if _, err := os.Stat(file); os.IsNotExist(err) {
```
stellarspot: This code can be written in a succinct way:
```
	_, err := os.Stat(fileName)
	return !os.IsNotExist(err)
```

Issue #27, opened 2018-08-09
Improve daemon logging


ldub:
Currently snet-daemon only logs blockchain interactions and RPC calls 

Logging facility should be improved to include
1. Service API metrics
1. Service performance metrics
1. Instrumentation/Debugging
1. Support for sending logs to third party hosting/analysis service

vsbogd:
Here is some proposal for the issue:
- at the moment [logrus](https://github.com/sirupsen/logrus) is used for logging 
- ```logrus``` has hooks to send some messages to the syslog or slack for example, new hooks can be added if it is needed
- to give client ability change logger settings these settings should be kept in external configuration file
- it is simpler to have one file for the whole snet-daemon configuration so logging settings should be added to the current JSON config
- to simplify configuration for the client we could introduce ```snet-daemon init --config <config_file>``` command to create ```<config_file>``` and print default config to it
- [logrus_mate](https://github.com/gogap/logrus_mate) could be used to load configuration and configure logger; logrus_mate can be configured to use JSON provider instead of default HOCON
- ```logrus_mate``` also has log rotation feature

pennachin:
Makes sense to me.

vsbogd:
Things which are done:
- init command added
- default configuration for logger added
- formatter type is configurable
- output type is configurable 
- log rotation supported
- log timestamps timezone is configurable

Things which are not done yet:
- support mail logrus hook as example of hook support
- add documentation

Improvement:
- use one value to configure logging timezone (at the moment one is used to configure logging timestamps and one to configure filename pattern)

vsbogd:
I would propose closing this issue:
- adding metrics to snet-daemon is a subject of issue https://github.com/singnet/snet-daemon/issues/30
- current logger configuration allows adding additional loggers for metrics if it is required
- support for sending logs to the 3rd party services is implemented, new services can be easily added

pennachin:
Agreed.

PR #41, opened 2018-09-24
Issue #27: add logger configuration

vsbogd:
Work on issue https://github.com/singnet/snet-daemon/issues/27.

What is done:
- add default configuration for the snet-daemon including logger configuration
- add logger module with InitLogger function to parse the configuration and initialize logger
- add wrapper for logger formatter to format logger timestamps using configured timezone
- add file writer which supports log rotation
- add unit tests for the logger initialization subsystem

What is not done:
- documentation for the logger configuration

### DIFF
--- a/logger/logger.go
+++ b/logger/logger.go
@@ -1,0 +1,19 @@
+package logger
+
+import (
+	"github.com/singnet/snet-daemon/config"
+	log "github.com/sirupsen/logrus"
+)
+
+func InitLogger() {
+	var err error
+	var logLevel log.Level
+	var settings = config.GetString(config.LogLevelKey)
+
+	logLevel, err = log.ParseLevel(settings)
+	if err != nil {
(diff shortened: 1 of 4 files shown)


logger/logger.go
```diff
@@ -0,0 +1,166 @@
...
+		return fmt.Errorf("Unable initialize log formatter, error: %v", err)
+	}
+	log.SetFormatter(formatter)
+
+	var output io.Writer
+	output, err = newOutputByConfig(config.sub(LogOutputKey))
+	if err != nil {
+		return fmt.Errorf("Unable initialize log output, error: %v", err)
+	}
+	log.SetOutput(output)
```
stellarspot: Should the output writer be closed on the app side in some way or it is handled by the logrus?

vsbogd:
Summary: the only thing we can to is to call Close() when log.Fatal() is called, this is the same thing which operating system does when application is closed by a signal or as result of os.Exit(). So we can add log.Fatal() handler but it seems there is no practical sense to do it.

PR #46, opened 2018-09-26
More elegant solution for viper subs with defaults

vsbogd:
Implement SubWithDefault method to replace viper.Sub(). viper.Sub() suffers from issue https://github.com/spf13/viper/issues/559. SubWithDefault method workarounds issue by adding default values to the sub configuration returned.

Fix logger initialization code accordingly. Add unit tests for new methods and fix old unit tests.

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -71,7 +71,6 @@ const (
 )
 
 var vip *viper.Viper
-var defaults *viper.Viper
 
 func init() {
 	var err error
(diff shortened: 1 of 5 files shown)


config/config.go
```diff
@@ -174,3 +172,22 @@ func GetDuration(key string) time.Duration {
 func GetBool(key string) bool {
 	return vip.GetBool(key)
 }
+
+// SubWithDefault returns sub-config by keys including configuration defaults
+// values. It returns nil if no such key. It is analog of the viper.Sub()
+// function. This is workaround for the issue
+// https://github.com/spf13/viper/issues/559
+func SubWithDefault(config *viper.Viper, key string) *viper.Viper {
+	var allSettingsByKey, ok = config.AllSettings()[strings.ToLower(key)]
```
stellarspot: Is it possible that a key with upper case characters were added to the config?

vsbogd: viper.Viper keeps all keys in low case internally, so it doesn't matter which case was used in key to put the value, when you get all the key/value map back all keys are in low case. For any Get() operation viper.Viper converts key to low case as well.

Issue #31, opened 2018-08-10
Replica Mode


ldub:
As part of the Service Arbiter Replica feature, we need the ability for an snet-daemon instance to be spun up as replica of another instance, providing the same service and authenticated to work on the same job contracts. This feature will enable:

* failover instances / reliability 
* per-region service endpoints
* load balancing
* automated scaling

vsbogd:
This issue consists of:
- https://github.com/singnet/snet-daemon/issues/72 to support MultiPartyEscrow payment channels
- https://github.com/singnet/snet-daemon/issues/63 to use etcd as shared payment channel storage

PR #54, opened 2018-10-02
Add gRPC interceptor stub to check gRPC escrow contract payment channel state

vsbogd:
Start of the work on issue https://github.com/singnet/snet-daemon/issues/31

As we are introducing new way of payment MultiPartyEscrow contract (see https://github.com/singnet/platform-contracts/pull/39) we need to support it in ```snet-daemon```. In order to do this I have modified gRPC interceptor to switch between two different ways of payment:
- job - an old one
- escrow - new way using escrow contract

Proposed API is described at [multiPartyEscrowApiChanges.md#daemon-api-changes](https://github.com/singnet/wiki/blob/master/multiPartyEscrowContract/multiPartyEscrowApiChanges.md#daemon-api-changes) page

What is done:
- move Job verification code to the ```blockchain/job.go```
- move MultiPartyEscrow payment channel verification code to the ```blockchain/escrow.go```
- add factory method in ```blockchain/interceptor.go``` to get either Job or Escrow payment verifier
- specify PaymentChannel storage API which will be used by snet-daemon to validate escrow contract payment channel state

Next steps:
- implement escrow contract payment channel verification logic

### DIFF
--- a/blockchain/escrow.go
+++ b/blockchain/escrow.go
@@ -1,0 +1,58 @@
+package blockchain
+
+import (
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/metadata"
+	"google.golang.org/grpc/status"
+	"math/big"
+)
+
+type PaymentChannelKey struct {
+	Id    *big.Int
+	Nonce *big.Int
+}
+
+type PaymentChannelState struct {
+	FullAmount       *big.Int
+	AuthorizedAmount *big.Int
+	PaymentSignature []byte
+}
+
+type PaymentChannelStorage interface {
+	Get(key *PaymentChannelKey) *PaymentChannelState
+	CompareAndSwap(key *PaymentChannelKey, prevState *PaymentChannelState, newState *PaymentChannelState) error
+}
+
+// escrowPaymentHandler implements paymentHandlerType interface
+type escrowPaymentHandler struct {
+	md      metadata.MD
+	storage PaymentChannelStorage
+}
+
+func newEscrowPaymentHandler() *escrowPaymentHandler {
+	return &escrowPaymentHandler{}
+}
+
+func (h *escrowPaymentHandler) validatePayment() error {
+	/*
+		id, err := getBigInt(h.md, PaymentChannelIdHeader)
+		if err != nil {
+			return err
+		}
+
+		nonce, err := getBigInt(h.md, PaymentChannelNonceHeader)
+		if err != nil {
+			return err
+		}
+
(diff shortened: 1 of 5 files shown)


blockchain/escrow.go
```diff
@@ -0,0 +1,58 @@
...
+
+type PaymentChannelKey struct {
+	Id    *big.Int
+	Nonce *big.Int
+}
+
+type PaymentChannelState struct {
+	FullAmount       *big.Int
+	AuthorizedAmount *big.Int
+	PaymentSignature []byte
```
vsbogd: + expiration date

blockchain/interceptors.go
```diff
@@ -16,29 +28,65 @@ func (p Processor) jobValidationInterceptor(srv interface{}, ss grpc.ServerStrea
...
+
+func getBigInt(md metadata.MD, key string) (*big.Int, error) {
+	array := md.Get(key)
+	if len(array) == 0 {
+		return nil, status.Errorf(codes.InvalidArgument, "missing \"%v\"", key)
 	}
 
-	return status.Errorf(codes.Unauthenticated, "job invocation not valid")
+	value := big.NewInt(0)
+	err := value.UnmarshalText([]byte(array[0]))
```
stellarspot: [hexutil.DecodeBig ](https://godoc.org/github.com/ethereum/go-ethereum/common/hexutil#DecodeBig) allows to decode a hex string to a big.Int
hexutil allows to decode a string to byte array as well.

blockchain/interceptors.go
```diff
@@ -16,29 +28,65 @@ func (p Processor) jobValidationInterceptor(srv interface{}, ss grpc.ServerStrea
...
 
-	if p.IsValidJobInvocation(jobAddressBytes, jobSignatureBytes) {
-		err := handler(srv, ss)
-		if err == nil {
-			p.CompleteJob(jobAddressBytes, jobSignatureBytes)
-		}
-		return err
+type paymentHandlerType interface {
+	validatePayment() error
+	completePayment(error) error
```
stellarspot: It would be good to have an interface where all functions take an error as an argument of neither of them.

blockchain/interceptors.go
vsbogd: Thanks, it seems that using ```-bin``` as key postfix means that value is binary and will be passed in ```base64``` format. And this convention is supported by all gRPC clients. I will test it and will use it for signature field.

Regarding this method it should decode string which contains decimal number. This format is used for incrementing ids such as channel_id and channel_nonce. It is not used in this code and not tested yet. I am going to test it and use it in next PR, so I would live it here if you don't mind.

PR #59, opened 2018-10-09
Implement escrow payment verification

vsbogd:
Working on issue https://github.com/singnet/snet-daemon/issues/31

This PR contains implementation of escrow contract payment channel verification algorithm. Following things are checked:
- channel exists and open
- signature is correct and provided by channel sender
- channel is not expired
- payment amount is not greater then channel amount
- payment amount increment is valid according pricing policy

After payment check service implementation is called. If service implementation returned result successful, then channel state is updated atomically.

This implementation relies on two interfaces:
- channel state storage - which should keep channel state and update it
- income validator - which validates that amount incremented correctly
Both interfaces are not implemented yet.

Implemented:
- MultiPartyEscrow contract payment verification algorithm
- unit tests

Next steps:
- implement channel storage using local database
- import new channels from blockchain
- implement income validation interface


blockchain/escrow.go
```diff
@@ -1,75 +1,289 @@
...
+
+	if payment.channel.FullAmount.Cmp(payment.amount) < 0 {
+		log.Warn("Not enough tokens on payment channel")
+		return status.Newf(codes.Unauthenticated, "not enough tokens on payment channel, channel amount: %v, payment amount: %v", payment.channel.FullAmount, payment.amount)
+	}
+
+	income := big.NewInt(0)
+	income.Sub(payment.amount, payment.channel.AuthorizedAmount)
+	err = h.incomeValidator.Validate(&IncomeData{Income: income, GrpcContext: payment.grpcContext})
+	if err != nil {
```
stellarspot: This check is unnecessary.

blockchain/interceptors.go
```diff
@@ -20,79 +22,183 @@ const (
...
+// Payment represents payment handler specific data which is validated
+// and used to complete payment.
+type Payment interface{}
+
+// PaymentHandler interface which is used by gRPC interceptor to get, validate
+// and complete payment. There are two payment handler implementations so far:
+// jobPaymentHandler and escrowPaymentHandler.
+type PaymentHandler interface {
+	// Payment extracts payment data from gRPC request context.
+	Payment(context *GrpcStreamContext) (payment Payment, err *status.Status)
```
stellarspot: "Payment" word is a noun. Does it have sense to use verb for the method name like "GetPayment"?

blockchain/interceptors.go
```diff
@@ -20,79 +22,183 @@ const (
...
+	log.Debug("Payment validated")
+
+	e := handler(srv, ss)
+	if e != nil {
+		log.WithError(e).Warn("gRPC handler returned error")
+		err = paymentHandler.CompleteAfterError(payment, e)
+		if err != nil {
+			return err.Err()
+		}
+		return e
```
stellarspot: If CompleteAfterError() succeeded does it necessary  to throw error after it?
If yes, could CompleteAfterError() always returns an error (new one or original) so we always returns it from the  CompleteAfterError() method call?

blockchain/interceptors.go
vsbogd: Go coding style recommends not using ```Get``` in such situations :-), see https://golang.org/doc/effective_go.html#Getters

blockchain/interceptors.go
vsbogd: Yes it is necessary, because ```e``` and ```err``` are two different errors here. ```CompleteAfterError``` is method which should cleanup payment state after call handler returns. If it succeeds we still should notify service caller that his request ended with error. If it ```CompleteAfterError``` returns error in turn we have a choice to return ```e``` or ```err```. Actually after thinking about it twice I think that it is better still return ```e```, but log ```err``` instead.

blockchain/escrow.go
```diff
@@ -1,75 +1,289 @@
...
+		&PaymentChannelData{
+			State:            payment.channel.State,
+			Sender:           payment.channel.Sender,
+			FullAmount:       payment.channel.FullAmount,
+			Expiration:       payment.channel.Expiration,
+			AuthorizedAmount: payment.amount,
+			Signature:        payment.signature,
+		},
+	)
+	if e != nil {
```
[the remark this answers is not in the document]
vsbogd: In this usage it doesn't matter as we are requiring that state should be the same and if CAS operation is not successful it doesn't matter why we should return error to caller. May be it make sense to change CAS API to return (bool, error) to make it less surprising.

PR #64, opened 2018-10-17
Add etcd storage

stellarspot:
* EtcdStorage which implements PaymentChannelStorage interface is added to escrow package
* simple EtcdStorage test which starts local etcd server is added into escrow folder
* EtcdClient endpoints are moved into EtcdClient config part.

### DIFF
--- a/escrow/etcd_storage.go
+++ b/escrow/etcd_storage.go
@@ -1,0 +1,121 @@
+package escrow
+
+import (
+	"bytes"
+	"encoding/gob"
+
+	"github.com/singnet/snet-daemon/etcddb"
+	"github.com/spf13/viper"
+)
+
+// EtcdStorage storage based on etcd
+type EtcdStorage struct {
+	client *etcddb.EtcdClient
+}
+
+// NewEtcdStorage create an etcd storage from Ciper config
+func NewEtcdStorage(vip *viper.Viper) (storage *EtcdStorage, err error) {
+	client, err := etcddb.NewEtcdClient(vip)
+
+	if err != nil {
+		return
+	}
+
+	storage = &EtcdStorage{client: client}
+	return
+}
+
+// Get gets value from etcd by key
+func (storage *EtcdStorage) Get(key *PaymentChannelKey) (state *PaymentChannelData, ok bool, err error) {
+
+	keyBytes, err := serialize(key)
+	if err != nil {
+		return
+	}
+
+	valueBytes, err := storage.client.Get(keyBytes)
+
+	if err != nil || len(valueBytes) == 0 {
+		return
+	}
+
+	state = &PaymentChannelData{}
+	err = deserialize(valueBytes, state)
+
+	if err != nil {
+		return
+	}
+
+	ok = true
+	return
+}
+
+// Put puts key and value to etcd
+func (storage *EtcdStorage) Put(key *PaymentChannelKey, channel *PaymentChannelData) (err error) {
+
+	keyBytes, err := serialize(key)
+	if err != nil {
+		return
+	}
+
+	valueBytes, err := serialize(channel)
+	if err != nil {
+		return
+	}
+
+	err = storage.client.Put(keyBytes, valueBytes)
+	return
+}
+
+// CompareAndSwap uses CAS operation to set a value
+func (storage *EtcdStorage) CompareAndSwap(
+	key *PaymentChannelKey,
+	prevState *PaymentChannelData,
+	newState *PaymentChannelData,
+) (ok bool, err error) {
+
+	keyBytes, err := serialize(key)
+	if err != nil {
+		return
+	}
+
+	expectBytes, err := serialize(prevState)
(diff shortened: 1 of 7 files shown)


escrow/etcd_storage.go
```diff
@@ -0,0 +1,121 @@
...
+func (storage *EtcdStorage) Get(key *PaymentChannelKey) (state *PaymentChannelData, ok bool, err error) {
+
+	keyBytes, err := serialize(key)
+	if err != nil {
+		return
+	}
+
+	valueBytes, err := storage.client.Get(keyBytes)
+
+	if err != nil || len(valueBytes) == 0 {
```
vsbogd: How do the caller know whether value is found or not? Get() method which is implemented here has ok flag which is false if key was not found. As far as I see EtcdClient.Get method returns nil when key is not found. Why len(valueBytes) is checked here?

escrow/etcd_storage.go
```diff
@@ -0,0 +1,121 @@
...
+	prevState *PaymentChannelData,
+	newState *PaymentChannelData,
+) (ok bool, err error) {
+
+	keyBytes, err := serialize(key)
+	if err != nil {
+		return
+	}
+
+	expectBytes, err := serialize(prevState)
```
vsbogd: As special case previous state can be nil to designate that value is absent in storage. Not sure this code handles situation correctly. Actually I am thinking about adding PutIfAbsent method instead of CAS(key, nil, state). What do you think?

escrow/etcd_storage_test.go
```diff
@@ -0,0 +1,215 @@
...
+	"github.com/singnet/snet-daemon/handler"
+	log "github.com/sirupsen/logrus"
+	"github.com/spf13/viper"
+	"github.com/stretchr/testify/assert"
+)
+
+var etcdPaymentHandler escrowPaymentHandler
+var etcdStorageMock etcdStorageMockType
+
+type etcdStorageMockType struct {
```
vsbogd: I would propose renaming it to ```cleanableEtcdStorage``` because ```etcdStorageMock``` means some fake etcd storage, but test itself tests real storage as far as I see. 

Solution which seems for me even better is to add some methods to cleanup storage using etcd API. 

I agree to postpone fixes till next PR.

stellarspot: I renamed the etcdStorageMock in the branch:
https://github.com/stellarspot/snet-daemon/commit/7d79313a742e6a9d0899f3b457b4bd90ad800cd3

etcddb/etcddb_client.go
```diff
@@ -29,30 +29,18 @@ type EtcdClient struct {
...
 		return
 	}
 
-	if conf == nil {
-		conf = &PaymentChannelStorageClientConf{
-			ConnectionTimeout: DefaultConnectionTimeout,
-			RequestTimeout:    DefaultRequestTimeout,
-		}
-	}
+	log.WithField("etcd client conf", fmt.Sprintf("%+v", conf)).Info()
```
vsbogd: Minor, but currently in code camelCase is used for log fields names, ideally I think name should be equal to the variable name to quickly find it in code

etcddb/etcddb_server.go
```diff
@@ -77,12 +72,9 @@ func getEtcdConf(conf *PaymentChannelStorageServerConf, cluster string) *embed.C
 		Host:   fmt.Sprintf("%s:%d", conf.Host, conf.PeerPort),
 	}
 
-	log.WithFields(log.Fields{
-		"etcd id":         conf.ID,
-		"client url":      clientURL,
-		"peer   url":      peerURL,
-		"initial cluster": cluster,
-	}).Info()
+	log.WithField("etcd server conf", fmt.Sprintf("%+v", conf)).Info()
```
vsbogd: Same minor thing as in https://github.com/singnet/snet-daemon/pull/64/files#r226319411

escrow/etcd_storage.go
stellarspot: I added the ok to the etcd client Get return value in the branch:
https://github.com/stellarspot/snet-daemon/commit/c417f27d50de9fa58cc14daf083cc33484455a34

escrow/etcd_storage.go
stellarspot: I filled an issue on it: https://github.com/singnet/snet-daemon/issues/68

etcddb/etcddb_server.go
stellarspot: Updated in the branch: https://github.com/stellarspot/snet-daemon/commit/1c512e6fc2dc9be24e8904a43ecc2bf48c2532b1

etcddb/etcddb_client.go
stellarspot: Updated in the branch: https://github.com/stellarspot/snet-daemon/commit/1c512e6fc2dc9be24e8904a43ecc2bf48c2532b1

PR #69, opened 2018-10-18
Use default config for etcd client/server from config package

stellarspot:
This fix adds :
* reading config package for default config
* some minor changes requested in the PR https://github.com/singnet/snet-daemon/pull/64


etcddb/payment_channel_storage.go
```diff
@@ -74,21 +74,24 @@ type PaymentChannelStorageServerConf struct {
...
+		Host:       "127.0.0.1",
+		ClientPort: 2379,
+		PeerPort:   2380,
+		Token:      "unique-token",
+		Cluster:    "storage-1=http://127.0.0.1:2380",
+		Enabled:    false,
 	}
 
-	if vip.Get(PaymentChannelStorageServerKey) == nil {
+	if !vip.InConfig(PaymentChannelStorageServerKey) {
```
vsbogd: It should be enough to add values into default config as you made in config/config.go . Default config is loaded as defaults to the viper and if user skipped some part of the configuration the default one will be returned. So I would propose remove manual defaults handling.

stellarspot: Viper UnmarshalKey(key) does not read values from defaults.
I have filled a separate issue on this change: #74

vsbogd: To make it more clear it actually reads but it suffers from issue which is the similar to https://github.com/spf13/viper/issues/559

Ok, lets merge it and continue discussion in the issue https://github.com/singnet/snet-daemon/issues/74

Issue #68, opened 2018-10-18
Investigate new values writing/reading in etcd


stellarspot:
See: https://github.com/singnet/snet-daemon/pull/64#discussion_r226307976

Foe example, it needs to investigate how to use CAS operation in etcd for the first time when a key value does not exist in etcd.

stellarspot:
It seems that there is no support in etcd Go API to run CAS with expect nil value:
https://github.com/etcd-io/etcd/issues/2880

The workaround can be to use etcd concurrency API to lock get/put operations while checking nil vallue:
https://godoc.org/github.com/coreos/etcd/clientv3/concurrency

stellarspot:
Fixed as part for the AtomicStorage API PutIfAbsent(key, value) method.
See https://github.com/singnet/snet-daemon/pull/78/commits/e885258306756e4967bc6106aa463f8ceece0a22

PR #73, opened 2018-10-19
#68 Update etcd client CAS to ise nil expect value

stellarspot:
See issue #68 
etcd client v3 API does not support using CAS with nil values.
The proposed solution uses etcd concurrency API to obtain a lock during get/put operation in CAS with nil expect value.


stellarspot:
Closed as PutIfAbsement is implemented in new pull request: #78

Issue #72, opened 2018-10-19
Support MultiPartyEscrow contract payment channels


vsbogd:
This is a part of issue https://github.com/singnet/snet-daemon/issues/31. I am adding this separate issue as issue https://github.com/singnet/snet-daemon/issues/31 to track the progress on MultiPartyEscrow (MPE) contract support. All MPE specifications can be found at https://github.com/singnet/wiki/tree/master/multiPartyEscrowContract

At the moment number of PRs are merged:
- PR https://github.com/singnet/snet-daemon/pull/54:
  - introduces new payment channel verification using MPE contracts
  - keeps old Job logic supported for the migration period
- PR https://github.com/singnet/snet-daemon/pull/59:
  - implements MultiPartyEscrow contract payment verification algorithm
- PR https://github.com/singnet/snet-daemon/pull/62 and PR https://github.com/singnet/snet-daemon/pull/70
  - implements logic of getting initial payment channel state from blockchain, putting it into storage and then working with storage

Things to be done:
- [x] return information about current payment channel state to the client according to https://github.com/singnet/wiki/blob/master/multiPartyEscrowContract/MultiPartyEscrow_stateless_client.md
- [x] implement income validator by getting price from Agent contract
- [x] support expiration time representation in block instead of time
- [x] add parameter to configure time prior to expiration time when daemon stops accepting payments from the channel
- [x] add snet-daemon command to claim money from the channel and update the channel state

vsbogd:
According to [MultiPartyEscrow stateles client design](https://github.com/singnet/wiki/blob/master/multiPartyEscrowContract/MultiPartyEscrow_stateless_client.md) service should return current channel state to the client which doesn't have it. Client needs the state in situation:
- after restart - client needs the latest state to start transactions but it cannot get it from blockchain because it authorized part of funds to withdrawing by service provider already;
- service provider withdrew part of the funds from the channel - this operation increments the channel nonce, so client should use new nonce to continue working with server.

It is suggested to add separate gRPC call for returning latest payment channel state of the client which requested it. gRPC interceptor should be modified to not validate the payment for the call.

vsbogd:
As Agent contract to be removed in future and at the moment there is no new implementation of price storage it is suggested to implement income validator which uses price from config for demonstration purposes.

vsbogd:
All coding is finished, but there are still some things to do:
- [ ] improve unit test coverage
- [ ] fix snet-daemon documentation

[...]

PR #79, opened 2018-10-22
Implement payment channel storage using atomic storage

vsbogd:
Refactoring to unblock work on issue https://github.com/singnet/snet-daemon/issues/72

What is done:
- move in-memory storage to separate file
- remove nonce from channel state storage key
- implement atomic storage API for in-memory storage
- implement payment channel storage using atomic storage API


vsbogd:
> I have a general question. Are there use cases which require to delete some data from the AtomicStorage?
> If yes, should AtomicStorage support some kind of Delete(key) method or just Put(key, nil) can be used?

Yes we have such usecases, but they are not implemented yet. I would vote for Delete(key) method as it has clear semantics.

PR #82, opened 2018-10-23
Add GetWithRange() method to etcd client

stellarspot:


etcddb/etcddb_client.go
```diff
@@ -81,6 +81,28 @@ func (client *EtcdClient) Get(key string) (value string, ok bool, err error) {
 	return
 }
 
+// GetWithRange gets all values where key is prefix
```
vsbogd: Am I right that prefix should be separated by dash from the key tail? If so it is worth to add it to documentation.

stellarspot: No, it doesn't. I used the dash just in test but ant prefix even without a separator is fine.

Etcd client v3 API allows to use [namespaces](https://godoc.org/github.com/coreos/etcd/clientv3/namespace) for keys, but it mostly used to separate keys for different applications which use the same etcd server.

stellarspot:
It is not possible to merge the pull request because of CircleCI is not triggered.
I reopen new one #84 instead of it.

PR #84, opened 2018-10-24
Add GetWithRange() method to etcd client

stellarspot:


etcddb/etcddb_client.go
```diff
@@ -81,6 +81,28 @@ func (client *EtcdClient) Get(key string) (value string, ok bool, err error) {
 	return
 }
 
+// GetWithRange gets all values where key is prefix
+func (client *EtcdClient) GetWithRange(key string) (values []string, ok bool, err error) {
```
vsbogd: I would propose renaming to GetByKeyPrefix

stellarspot: Renamed the method to GetByKeyPrefix.

PR #86, opened 2018-10-24
Use price per call from config by IncomeValidator

stellarspot:


escrow/income.go
```diff
@@ -41,7 +45,15 @@ func NewIncomeValidator(processor *blockchain.Processor) (validator IncomeValida
...
-func (validator *incomeValidator) Validate(*IncomeData) (err *status.Status) {
-	// TODO: implement
-	return nil
+func (validator *incomeValidator) Validate(data *IncomeData) (err *status.Status) {
+
+	price := config.GetBigInt(config.PricePerCallKey)
+
+	if data.Income.Cmp(price) != 0 {
+		msg := fmt.Sprintf("income %d does not equal to price %d", data.Income, price)
+		err = status.New(codes.Internal, msg)
```
vsbogd: Please return codes.Unauthenticated, it is a code which is used to designate that payment cannot be accepted: https://github.com/singnet/snet-daemon/blob/a14ee6266ca979cbdb8b422fedaab337ac217258/escrow/escrow.go#L264

vsbogd:
This is an implementation of simple income validator task from issue https://github.com/singnet/snet-daemon/issues/72

escrow/income.go
stellarspot: Updated error code to Unauthenticated.

PR #89, opened 2018-10-25
Implement channel claim part 2

vsbogd:
Work on issue https://github.com/singnet/snet-daemon/issues/72

What is done:
- added code to call MultiPartyEscrow.channelClaim 
- added flags to pass timeout and sendBack parameters

To be done:
- keep list of all payments which were sent to blockchain in etcd
- atomically add payment to etcd and change channel state
- remove payment from etcd when transaction is committed to blockchain

### DIFF
--- a/blockchain/escrow.go
+++ b/blockchain/escrow.go
@@ -1,0 +1,59 @@
+package blockchain
+
+import (
+	"context"
+	"fmt"
+	"math/big"
+	"time"
+
+	"github.com/ethereum/go-ethereum/accounts/abi/bind"
+	"github.com/ethereum/go-ethereum/common"
+	log "github.com/sirupsen/logrus"
+)
+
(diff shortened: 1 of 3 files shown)


blockchain/escrow.go
```diff
@@ -0,0 +1,59 @@
...
+	if err != nil {
+		log.WithError(err).Error("Error submitting transaction to claim funds from channel")
+		return fmt.Errorf("Error submitting transaction to claim funds from channel: %v", err)
+	}
+
+	log.Info("Transaction sent, waiting for %v till transaction is committed", timeout)
+	endTime := time.Now().Add(timeout)
+	isPending := true
+	for {
+		_, isPending, _ = processor.ethClient.TransactionByHash(context.Background(), txn.Hash())
```
stellarspot: It is also possible  that isPending is false but returns error is not nil.
In this case the transaction should not be treated as successful.

PR #92, opened 2018-10-25
Implement lazy dependency injection

vsbogd:
Refactoring to initialize components only when they are used.

### DIFF
--- a/snetd/cmd/components.go
+++ b/snetd/cmd/components.go
@@ -27,12 +27,14 @@ type Components struct {
 	paymentChannelStateService *escrow.PaymentChannelStateService
 }
 
-func InitComponents(cmd *cobra.Command) (components *Components, err error) {
+func InitComponents(cmd *cobra.Command) (components *Components) {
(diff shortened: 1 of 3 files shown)


snetd/cmd/components.go
```diff
@@ -78,119 +64,144 @@ func isFileExist(fileName string) bool {
...
-	if components.etcdClient != nil {
-		components.etcdClient.Close()
+
+	var delegateStorage escrow.AtomicStorage
+	if config.GetString(config.PaymentChannelStorageTypeKey) == "etcd" {
+		delegateStorage = components.EtcdClient()
+	} else {
+		delegateStorage = escrow.NewMemStorage()
 	}
 	if components.etcdServer != nil {
```
stellarspot: Probably some conflicts during merging.
EtcdServer shouldn't be closed after PaymentChannelStorage creation.

vsbogd: Thanks!

PR #101, opened 2018-10-26
Clean etcd tests work directories

stellarspot:
* **data_dir** property is added to the EtcdServer configuration file
* gunit is used in tests to run setup and teardown methods
* test data dirs are cleaned

### DIFF
--- a/etcddb/etcddb_test.go
+++ b/etcddb/etcddb_test.go
@@ -2,6 +2,9 @@ package etcddb
 
 import (
 	"fmt"
+	"io/ioutil"
+	"os"
+	"strings"
 	"testing"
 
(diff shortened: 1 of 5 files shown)


etcddb/etcddb_test.go
```diff
@@ -2,14 +2,31 @@ package etcddb
 
 import (
 	"fmt"
+	"os"
 	"testing"
 
+	"github.com/smartystreets/gunit"
```
vsbogd: I would suggest using ```github.com/stretchr/testify/suite``` to not introduce new library.

stellarspot: Updated the test to use testify/suite instead of gunit

Issue #107, opened 2018-10-30
Daemon claim command should not try starting etcd server


vsbogd:
At the moment ```snet-daemon claim``` command tries to start etcd server but in fact it is not required because etcd client is enough.

PR #122, opened 2018-11-06
Daemon claim command should not try starting etcd server #107

stellarspot:


stellarspot:
I updated the to call etcd server directly in the serve command instead of using RunMode flag.

PR #123, opened 2018-11-07
Pass etcd logs to logrus

stellarspot:
* add "log_level" field to etcd server config
* Implement capnslog to logrus formatter
* pass etcd logs to logrus

### DIFF
--- a/etcddb/etcddb_conf.go
+++ b/etcddb/etcddb_conf.go
@@ -1,7 +1,6 @@
 package etcddb
 
 import (
-	"fmt"
 	"strings"
 	"time"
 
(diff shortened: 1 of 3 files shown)


etcddb/etcddb_conf.go
```diff
@@ -84,5 +88,66 @@ func GetEtcdServerConf(vip *viper.Viper) (conf *EtcdServerConf, err error) {
...
+// with methods Format and Flush
+type capnslogToLogrusLogFormatter struct {
+}
+
+func (formatter *capnslogToLogrusLogFormatter) Format(pkg string, level capnslog.LogLevel,
+	depth int, entries ...interface{}) {
+
+	l := log.WithFields(log.Fields{
+		"pkg": pkg,
+		"msg": fmt.Sprint(entries...),
```
vsbogd: It seems more logical to format entries into the string and use as log message, for instance:
```go
msg := fmt.Sprintf(entries...)
...
l.Info(msg)
```

Issue #113, opened 2018-11-01
Get payment channel details


raamb:
Add an option to the client to get channel state from treasurer server so that service provider can decide to claim or not

PR #129, opened 2018-11-14
Add command to list channels

vsbogd:
Fix for issue https://github.com/singnet/snet-daemon/issues/113

New command ```snetd list channels``` is added. This command gets all of the channels from PaymentChannelService and prints them in console.

### DIFF
--- a/snetd/cmd/list_channels.go
+++ b/snetd/cmd/list_channels.go
@@ -1,6 +1,8 @@
 package cmd
 
 import (
+	"fmt"
+
 	"github.com/spf13/cobra"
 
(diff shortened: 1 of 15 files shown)


snetd/cmd/list_channels.go
```diff
@@ -0,0 +1,47 @@
...
+}
+
+func (command *listChannelsCommand) Run() (err error) {
+	channels, err := command.channelService.ListChannels()
+	if err != nil {
+		return
+	}
+
+	if len(channels) == 0 {
+		fmt.Println("no channels in shared storage")
```
stellarspot: It is possible to return from the zero channels length check.

PR #133, opened 2018-11-19
Support for new Registry and MPE Changes

Dinesh2521:
- Removes all the references to Agent and job contracts
- Read the metadata from the registry contract and update the configurations

@vsbogd @astroseger @raamb


blockchain/escrow.go
```diff
@@ -20,6 +20,11 @@ func (processor *Processor) ClaimFundsFromChannel(timeout time.Duration, channel
 		"isSendBack": sendBack,
 	})
 
+	v, r, s, err := ParseSignature(signature)
+	if err != nil {
+		return
```
raamb: We should log the error and return it here

blockchain/ethereum_test_util.go
```diff
@@ -51,8 +51,8 @@ func (env *SimulatedEthereumEnvironment) MpeDeposit(from *bind.TransactOpts, amo
 	return env
 }
 
-func (env *SimulatedEthereumEnvironment) MpeOpenChannel(from *bind.TransactOpts, to *bind.TransactOpts, amount int64, expiration int64, groupId int64) *SimulatedEthereumEnvironment {
-	_, err := env.MultiPartyEscrow.OpenChannel(EstimateGas(from), to.From, big.NewInt(amount), big.NewInt(expiration), big.NewInt(groupId))
+func (env *SimulatedEthereumEnvironment) MpeOpenChannel(from *bind.TransactOpts, to *bind.TransactOpts, amount int64, expiration int64, groupId [32]byte) *SimulatedEthereumEnvironment {
+	_, err := env.MultiPartyEscrow.OpenChannel(EstimateGas(from), to.From, big.NewInt(amount), big.NewInt(expiration), groupId, to.From)
 	if err != nil {
```
raamb: Please validate the OpenChannel call against the contract. The signature I see in the contract is openChannel(address  recipient, uint256 value, uint256 expiration, bytes32 groupId, address signer)

ipfsutils/serviceMetadata.go
```diff
@@ -0,0 +1,98 @@
...
+	groupName := GetDaemonGroupName()
+	for _, group := range metaData.Groups {
+		if strings.Compare(groupName, group.GroupName) == 0 {
+			paymentAddress = group.PaymentAddress
+		}
+	}
+	return paymentAddress
+}
+
+func StringToBytes32(str string) [32]byte {
```
raamb: Seem like the same function as in Blockchain/Utils. Lets reuse it

ipfsutils/ipfsutils.go
```diff
@@ -0,0 +1,41 @@
...
+	jsondata := string(blob)
+	re := regexp.MustCompile("\\n")
+	jsondata = re.ReplaceAllString(jsondata, " ")
+	cid.Close()
+	return jsondata
+}
+
+func GetIpfsShell() *shell.Shell {
+	//Read from Configuration file
+	sh := shell.NewShell("http://localhost:5002/")
```
vsbogd: Configuration IPFS endpoint should be used here

snetd/cmd/serve.go
```diff
@@ -118,7 +118,7 @@ func newDaemon(components *Components) (daemon, error) {
 
 func (d daemon) start() {
 
-	d.blockProc.StartLoop()
+	//d.blockProc.StartLoop()
```
vsbogd: Can be removed

config/config.go
```diff
@@ -46,6 +50,11 @@ const (
 	defaultConfigJson string = `
 {
 	"auto_ssl_cache_dir": ".certs",
+    "ipfs": {
```
vsbogd: I would suggest replace it by string "http://loclahost:5051" to simplify processing

ipfsutils/serviceMetadata.go
```diff
@@ -0,0 +1,98 @@
...
+	metaData = new(service_metadata)
+	json.Unmarshal([]byte(jsondata), &metaData)
+}
+
+func SetServiceMetaDataThroughJSON(jsondata string) {
+	metaData = new(service_metadata)
+	json.Unmarshal([]byte(jsondata), &metaData)
+}
+
+func GetPaymentExpirationThreshold() int {
```
vsbogd: Should be used in ChannelPaymentValidator.Validate()

ipfsutils/serviceMetadata.go
```diff
@@ -0,0 +1,98 @@
...
+	"github.com/singnet/snet-daemon/config"
+	"math/big"
+	"strings"
+)
+
+type service_metadata struct {
+	Version                    int    `json:"version"`
+	DisplayName                string `json:"display_name"`
+	Encoding                   string `json:"encoding"`
+	ServiceType                string `json:"service_type"`
```
vsbogd: Encoding and ServiceType should be used in code as well, I think it is ok to postpone it but then we need to raise new issue or keep previous issue opened.

ipfsutils/serviceMetadata.go
```diff
@@ -0,0 +1,98 @@
...
+	Endpoints []struct {
+		GroupName string `json:"group_name"`
+		Endpoint  string `json:"endpoint"`
+	} `json:"endpoints"`
+	DeamonReplicaGroupID string
+	DeamonGroupName      string
+	DaemonEndPoint       string
+}
+
+var metaData *service_metadata
```
vsbogd: Refactoring proposal (discussable and definitely can be postponed).

I would have ```service_metadata``` instance as a component instead of making global variable. It can be instantiated as a component in ```components.go``` and injected as dependency into all components which require it. It allows:
- having dependency graph in explicit form in ```components.go```; at the moment metadata is implicitly initialized in blockchain component and if there is a component which requires metadata but not requires blockchain it is not obvious that it should be initialized after blockchain;
- passing custom formed objects as parameters of the components in unit tests instead of setting global settings;

blockchain/ethereum_test_util.go
```diff
@@ -51,8 +51,8 @@ func (env *SimulatedEthereumEnvironment) MpeDeposit(from *bind.TransactOpts, amo
 	return env
 }
 
-func (env *SimulatedEthereumEnvironment) MpeOpenChannel(from *bind.TransactOpts, to *bind.TransactOpts, amount int64, expiration int64, groupId int64) *SimulatedEthereumEnvironment {
-	_, err := env.MultiPartyEscrow.OpenChannel(EstimateGas(from), to.From, big.NewInt(amount), big.NewInt(expiration), big.NewInt(groupId))
+func (env *SimulatedEthereumEnvironment) MpeOpenChannel(from *bind.TransactOpts, to *bind.TransactOpts, amount int64, expiration int64, groupId [32]byte) *SimulatedEthereumEnvironment {
+	_, err := env.MultiPartyEscrow.OpenChannel(EstimateGas(from), to.From, big.NewInt(amount), big.NewInt(expiration), groupId, to.From)
```
vsbogd: It is unclear because ```from``` and ```to``` parameters of bind.TransactOpts type which has an address in field From, so it should be:
```suggestion
	_, err := env.MultiPartyEscrow.OpenChannel(EstimateGas(from), to.From, big.NewInt(amount), big.NewInt(expiration), groupId, from.From)
```

ipfsutils/ipfsutils.go
```diff
@@ -0,0 +1,41 @@
...
+	if err != nil {
+		log.WithError(err).Panic("error executing the cat command in ipfs")
+	}
+
+	blob, err := ioutil.ReadAll(cid)
+	if err != nil {
+		log.WithError(err).Panic("error: in Reading the meta data file %s", err)
+
+	}
+	log.Debug(string(blob))
```
vsbogd:
```suggestion
	log.WithField("hash", hash).WithField("blob", string(blob)).Debug("Blob of IPFS file with hash")
```

vsbogd:
Please remove  blockchain/multi_party_escrow.go, blockchain/registry.go and  blockchain/singularity_net_token.go again

escrow/payment_channel_storage_test.go
```diff
@@ -144,7 +144,7 @@ func (suite *BlockchainChannelReaderSuite) TestGetChannelState() {
 
 func (suite *BlockchainChannelReaderSuite) TestGetChannelStateIncorrectGroupId() {
 	reader := suite.reader
-	reader.replicaGroupID = func() (*big.Int, error) { return big.NewInt(321), nil }
+	reader.replicaGroupID = func() ([32]byte, error) { return [32]byte{123}, nil }
```
vsbogd:
```suggestion
	reader.replicaGroupID = func() ([32]byte, error) { return [32]byte{321}, nil }
```

escrow/validation.go
```diff
@@ -81,10 +78,11 @@ func getSignerAddressFromPayment(payment *Payment) (signer *common.Address, err
 
 	signer, err = getSignerAddressFromMessage(message, payment.Signature)
 	if err != nil {
-		return
+		log.WithError(err)
```
vsbogd:
```suggestion
		log.WithField("payment", payment).WithError(err).Error("Cannot get signer from payment")
```

PR #140, opened 2018-11-21
Daemon configurations required are added on to the default config 

anandrgit:
Daemon configurations required are added on to the default config


config/config.go
```diff
@@ -41,27 +39,28 @@ const (
...
 	PaymentChannelStorageServerKey      = "PAYMENT_CHANNEL_STORAGE_SERVER"
 
 	defaultConfigJson string = `
 {
-	"auto_ssl_cache_dir": ".certs",
 	"auto_ssl_domain": "",
 	"blockchain_enabled": true,
 	"daemon_listening_port": 5000,
 	"daemon_type": "grpc",
+    "daemon_end_point": "http://localhost:8080",
```
vsbogd: Doesn't correspond to daemon_listening_port two lines above.

BTW, at the moment we have two properties to configure essentially same thing: ```daemon_listening_port``` and ```daemon_end_point```. I can imaging setup where these two things are not the same but it can be surprising for users. Don't have idea how to fix it properly so far. One of the options don't support such setups.

config/config.go
```diff
@@ -41,27 +39,28 @@ const (
...
 	"db_path": "snetd.db",
 	"ethereum_json_rpc_endpoint": "http://127.0.0.1:8545",
-    "IPFS_END_POINT": "http://localhost:5002/",
 	"hdwallet_index": 0,
 	"hdwallet_mnemonic": "",
+    "ipfs_end_point": "http://localhost:5002/",
```
vsbogd: Please pay attention on spaces :-) Probably I have used unusual for Golang formatting in this JSON, but I would prefer keep it consistent and fix in separate PR if it is bothering.

config/config.go
```diff
@@ -41,27 +39,28 @@ const (
...
-    "IPFS_END_POINT": "http://localhost:5002/",
 	"hdwallet_index": 0,
 	"hdwallet_mnemonic": "",
+    "ipfs_end_point": "http://localhost:5002/", 
+    "organization_name": "exampleorganization", 
 	"price_per_call": 10,
 	"passthrough_enabled": false,
 	"poll_sleep": "5s",
+    "registry_address_key": "0x4e74fefa82e83e0964f0d9f53c68e03f7298a8b2",
+    "service_name": "dnndisplay",
```
vsbogd: Minor, but I would expect something like ```exampleservice``` here.

config/config.go
vsbogd: Opened issue https://github.com/singnet/snet-daemon/issues/141  to discuss it

vsbogd:
General comment: it is better to raise non coupled changes using separate branch and separate pull request. It will simplify review and sometimes change will be merged faster. For instance last commit can be merged as separate PR, but should wait for review ended in this PR.

.gitignore
```diff
@@ -5,6 +5,7 @@ snetd.db
 snetd.config.json
 /vendor/
 /build/
+Gopkg.lock
```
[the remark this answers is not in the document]
anandrgit: reverted back this , ( this was coming up in the diff and hence tried to remove it)

PR #143, opened 2018-11-23
Changed to the snet-deamon README.md ( Removed references to Agent and Job) and added some details on Treasure Server. 

anandrgit:


anandrgit:
@vsbogd @astroseger , Request your review on this request.

README.md
```diff
@@ -74,7 +91,7 @@ Available Commands:
   help        Help about any command
   init        Write default configuration to file
   list        List channels, claims in progress, etc
-  serve       
+  serve       (default) option
```
vsbogd: 1. Text above is actually copied from ```./build/snetd-linux-amd64 --help``` output, so I would suggest adding description for serve command instead. 
2. May be it makes sense to rename ```serve``` to ```start``` command as it should be more obvious.
3. I think it is not required to keep ```serve``` (or ```start```) as default behavior, I think now it is time when we can replace all usages by explicit ```snetd-linux-amd64 start``` especially if it is confusing.

README.md
```diff
@@ -3,6 +3,17 @@
 [![CircleCI](https://circleci.com/gh/singnet/snet-daemon.svg?style=svg)](https://circleci.com/gh/singnet/snet-daemon)
 
 SingularityNET Daemon
+The daemon is the adapter with which an otherwise SingularityNET-unaware service implementation can be exposed to the SingularityNET platform. It is designed to be deployed as a sidecar proxy alongside the service on a given host.
+The daemon abstracts the blockchain components away from the clients.
+The SNET Daemon interacts with the Multi Party Escrow to facilitate authorization and payment for services and acts as a passthrough for making API calls to the service.The daemon is the endpoint a client will submit requests to, and they are then passed to the service after validation by the daemon.
+
+
+## Treasure Server
```
vsbogd: I think it is misleading to talk about "treasure server" when there is no such server yet. What we have is a command line tool to get a payment, update the channel state and write transaction to the blockchain. I would say at the moment it is a "channel claim cli".

astroseger: I would still propose to call it at least treasurer server logic. Otherwise people will not understand.

README.md
raamb: Lets track the default behaviour in a separate issue that I will create now. From a readme perspective lets release this now as it reflects current behaviour and fix it along with the issue

README.md
raamb: Agree with VItaly, lets call it "Channel Claim command" for now

anandrgit: fixed

README.md
anandrgit: for now , re worded to "Is the default option which starts the Daemon"

vsbogd: Sure, I didn't mean that 2 and 3 should be done in this PR.
Thanks, @raamb, for raising issue on this.

Issue #145, opened 2018-11-23
Call payment.CompleteAfterError() even if panic() happens


vsbogd:
See: 
https://github.com/singnet/snet-daemon/blob/82ae1ca8cf397b1335c866b136919d5f5ff49fa9/handler/interceptors.go#L109

This call should be deferred and called even when panic() happened  during handler call.

raamb:
@stellarspot can we close this issue? Your PR #154 resolves this issue I believe.

vsbogd:
@raamb , it is still not tested, no unit test and no manual test, I would close it after adding unit test.

PR #154, opened 2018-11-28
#145 Call payment.CompleteAfterError() even if panic() happens

stellarspot:
- paymentHandler.CompleteAfterError() method is deferred
- panic mode after handler(...) call is handler

### DIFF
--- a/handler/interceptors.go
+++ b/handler/interceptors.go
@@ -95,13 +95,6 @@ func (interceptor *paymentValidationInterceptor) intercept(srv interface{}, ss g
 	if err != nil {
 		return err.Err()
 	}
-	log.WithField("payment", payment).Debug("New payment received")
-
-	err = paymentHandler.Validate(payment)
-	if err != nil {
-		return err.Err()
-	}
-	log.Debug("Payment validated")
 
 	handlerSucceed := false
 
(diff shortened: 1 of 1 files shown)


handler/interceptors.go
```diff
@@ -103,13 +103,28 @@ func (interceptor *paymentValidationInterceptor) intercept(srv interface{}, ss g
...
+				err = paymentHandler.CompleteAfterError(payment, e)
+				if err != nil {
+					e = err.Err()
+				}
+			}
+		}
+	}()
+
+	e = handler(srv, ss)
+	handlerSucceed = err != nil
```
vsbogd: Should it be?
```suggestion
	handlerSucceed = e != nil
```
Or we could move ```handlerSucceed = true``` to the line 130

handler/interceptors.go
```diff
@@ -103,13 +103,28 @@ func (interceptor *paymentValidationInterceptor) intercept(srv interface{}, ss g
 	}
 	log.Debug("Payment validated")
 
-	e := handler(srv, ss)
+	handlerSucceed := false
+
+	defer func() {
```
vsbogd: I would suggest moving ```defer``` call to line 99, as it should also be called in case of panic or error within Validate() method.

handler/interceptors.go
stellarspot: Yes, it should check e. I have updated this according to the suggestion.

handler/interceptors.go
stellarspot: Updated.

PR #156, opened 2018-11-30
Remove database usage

vsbogd:
Remove old database usages as it is replaced by etcd


config/config.go
```diff
@@ -47,7 +46,6 @@ const (
 	"daemon_listening_port": 8080,
 	"daemon_type": "grpc",
 	"daemon_end_point": "127.0.0.1:8080",
-	"db_path": "snetd.db",
```
stellarspot: It looks like 'payment_channel_storage_type' in config file is not necessary anymore.

vsbogd: It is still possible to use in-memory storage

Issue #158, opened 2018-11-30
Segfault on starting daemon


vsbogd:
```newDaemon()``` function parses port from ```daemon_end_point``` property and returns error even when port is parsed correctly. Then in line 94 (see code below) nil is returned because ```Wrap()``` is called around err which is nil. ```Wrap()``` returns nil in such case so ```newDaemon()``` returns no error but daemon is not initialized correctly. In particular ```daemon.lis``` is nil which lead to segfault.

Source code: https://github.com/singnet/snet-daemon/blob/a5205d1a1090b53f6f421730fa285d9c1b5252bd/snetd/cmd/serve.go#L91-L95

Error log:
```
...
time="2018-11-30T07:26:10Z" level=info msg="daemon_end_point: http://localhost:8090"
...
time="2018-11-30T07:26:10Z" level=debug msg="starting daemon"
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x18 pc=0xf4308e]

goroutine 349 [running]:
github.com/singnet/snet-daemon/vendor/github.com/soheilhy/cmux.(*cMux).Serve(0xc420432640, 0x0, 0x0)
	/root/singnet/src/github.com/singnet/snet-daemon/vendor/github.com/soheilhy/cmux/cmux.go:162 +0x7e
created by github.com/singnet/snet-daemon/snetd/cmd.daemon.start
	/root/singnet/src/github.com/singnet/snet-daemon/snetd/cmd/serve.go:219 +0x668
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x28 pc=0xf46412]
```

vsbogd:
Fixed by #159 and #161

PR #161, opened 2018-11-30
#Issue 158 

anandrgit:
1) Correct the typo on the variable used to check error
2) Deriving the port number when the daemon end point is of any format below
    http://localhost:8080
    https://localhost:8080
    localhost:8080
    127.0.0.1:8080
https://github.com/singnet/snet-daemon/issues/158


anandrgit:
@vsbogd   request your review on this

Issue #151, opened 2018-11-28
Document snet-daemon configuration


vsbogd:
This is to document daemon configuration in snet-daemon/README.md file.
See issue https://github.com/singnet/wiki/issues/32.

PR #166, opened 2018-12-04
Update daemon configuration documentation

vsbogd:
Update snet-daemon configuration documentation to fix https://github.com/singnet/snet-daemon/issues/151
and unblock https://github.com/singnet/wiki/issues/32.

what is done:
- documentation updated
- property names are converted to the same case (lowcase)
- defaults for ```private_key```  and ```auto_ssl_cache_dir``` properties are added
- legacy ```daemon_listening_port``` property is removed

### DIFF
--- a/README.md
+++ b/README.md
@@ -119,96 +119,96 @@ configuration properties can be set using configuration file.
 These properties you should usually change before starting daemon for the first
 time.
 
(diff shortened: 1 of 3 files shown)


vsbogd:
@ferrouswheel could you please take a look

vsbogd:
I don't really like how it looks like, one of the options is to move the configuration description into ```config/README.md```. May be list will be better then headers.

PR #174, opened 2018-12-14
Support contract Changes on Using Organization Id and Service Id 

anandrgit:
Support contract Changes on Using Organization Id and Service Id instead of Organization name and Service Name.
We will need to clean up the daemon configuration generation across all wikis / README.md files to reflect this correctly.


anandrgit:
The build will fail as the latest contract is not available for reference at 0.2.4

vsbogd:
Looks good to me, as https://github.com/singnet/platform-contracts/pull/85 is merged already, could you please replace release version at https://github.com/singnet/snet-daemon/blob/e9ff650af297219ef2cafacf2ce88d7be6ddca29/resources/blockchain/package.json#L6
to make it buildable

Issue #110, opened 2018-11-01
Capture usage metrics on the Daemon side


raamb:
Capture usage metrics on the Daemon side - We need to provide a mechanism to opt-in / opt-out for the service provider

raamb:
#178 adds this functionality

raamb:
For monitoring purpose , every incoming request and the outgoing response will be intercepted and will published on to a monitoring service end point provided.
Monitoring interceptor will be the first interceptor ( even requests rejected due to rate limiting will also need to be tracked)
Based on these stats published corresponding metrics/reporting can be extracted.
Metrics collection is enabled by default in configuration. Service Provider can disable the metrics anytime,if they do not want to collect any metrics.

PR #178, opened 2018-12-26
Issue #110 : Daemon Monitoring -  Capture Request and Response details for reporting

anandrgit:
For monitoring purpose , every incoming request and the outgoing response will be intercepted and will published on to a monitoring service end point provided.
Monitoring interceptor will be the first interceptor ( even requests rejected due to rate limiting will also need to be tracked) 
Based on these stats published corresponding metrics/reporting can be extracted.
Metrics collection is enabled by default  in configuration. Service Provider can disable the metrics anytime,if they do not want to collect any metrics.


anandrgit:
hi @vsbogd  request your review on this PR 
We did have one round reviewed with raam and @Dinesh2521

Issue #176, opened 2018-12-18
check hash of files from IPFS


astroseger:
We should check hash of files from IPFS. Otherwise we can be attacked if ipfs client is compromised.

raamb:
PR #182 addresses this issue

PR #182, opened 2019-01-03
check hash of files from IPFS (fix for issue #176)

Dinesh2521:
@astroseger @vsbogd @raamb @anandrgit 
this fix involves generating the IPFS multihash from the IPFS provided metadata and validate against the requested IPFS metadata using Hash. Both generated and requested hashes must match to consider the metadata is valid and secure.

Can you review the fix and provide your comments

### DIFF
--- a/Gopkg.toml
+++ b/Gopkg.toml
@@ -69,3 +69,7 @@
 [[constraint]]
   name = "github.com/golang/protobuf"
   version = "1.2.0"
+
+[[constraint]]
+  name = "github.com/ipfs/go-ipfs-api"
+  branch = "master"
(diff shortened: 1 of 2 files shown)


Gopkg.toml
```diff
@@ -69,3 +69,7 @@
 [[constraint]]
   name = "github.com/golang/protobuf"
   version = "1.2.0"
+
+[[constraint]]
+  name = "github.com/ipfs/go-ipfs-api"
+  branch = "master"
```
vsbogd: See https://github.com/singnet/snet-daemon/pull/183/files#r245478941

Issue #170, opened 2018-12-06
Daemon hangs if no service metadata in IPFS


vsbogd:
When daemon started it tries to load service configuration reading service metadata from IPFS. If IPFS doesn't has such file it doesn't return error but returns empty file instead. Daemon hangs on start if metadata is empty.

End of the log:
```log
NFO[0001] Starting listening port                       port=8080
DEBU[0001] Before Formatting                             metadataHash="ipfs://QmTsNJzfPuK69J85fNP8JzHMT3gKTnarXpdaYuRJ2SkWaW\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00\x00"
DEBU[0001] After Formatting                              metadataUri=QmTsNJzfPuK69J85fNP8JzHMT3gKTnarXpdaYuRJ2SkWaW
DEBU[0001] Hash Used to retrieve from IPFS               hash=QmTsNJzfPuK69J85fNP8JzHMT3gKTnarXpdaYuRJ2SkWaW
```

raamb:
#183 adds a timeout to ipfs reads and resolves the hang issue

PR #183, opened 2019-01-03
fix for issue when Daemon hangs if no service metadata in IPFS (issue #170)

Dinesh2521:
@vsbogd @raamb @anandrgit 
By default, IPFS reads/writes has time out for more than a minute. If there is no file found in IPFS, then the shell keeps waiting for the file until timeout occurs, during which daemon looks hanged. 
A configurable timeout has been added (defaulted to 30 sec) to Daemon to control the timeout duration to reduce effects of empty IPFS or non existence of files in IPFS. if in case of a real empty IPFS file  returned, then unmarshalling to metadata fails and eventually daemon will not run anymore.

Please review and let us know you comments/suggestions.

### DIFF
--- a/Gopkg.toml
+++ b/Gopkg.toml
@@ -69,3 +69,7 @@
 [[constraint]]
   name = "github.com/golang/protobuf"
   version = "1.2.0"
+
+[[constraint]]
+  name = "github.com/ipfs/go-ipfs-api"
+  branch = "master"
--- a/README.md
+++ b/README.md
@@ -234,6 +234,7 @@ see [rate limiting configuration](./ratelimit/README.md)
 
 * **heartbeat_svc_end_point** (optional; default: `""`) - It must be a valid URL. if it is empty, then service state always assumed as SERVING, and same will be wrapped in Daemon Heartbeat. see [daemon heartbeats configuration](./metrics/README.md)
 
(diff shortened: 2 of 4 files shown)


Gopkg.toml
```diff
@@ -69,3 +69,7 @@
 [[constraint]]
   name = "github.com/golang/protobuf"
   version = "1.2.0"
+
+[[constraint]]
+  name = "github.com/ipfs/go-ipfs-api"
+  branch = "master"
```
vsbogd: Is it possible to stick to some specific release version? Sticking to master can lead to future build is broken because of incompatible change in IPFS master branch.

Dinesh2521: shell.OnlyHash(true), this option is available only in the master build. This functionality is new and doesn't available in release versions yet. 

there is no other means to generate the IPFS hash on the client side using plain content, because IPFS hash is not only the hash of the content, it is the combination of multiple hashes, thats why it is called multihash. IPFS gateway determines this hash when it splits the contents and sends to nodes. each piece hash is calculated separately. And IPFS hash is the Hash Addition of all the individual hashes on different nodes. This final used as content ID (cid), which we used as file address in IPFS.

vsbogd: ok, then

PR #184, opened 2019-01-09
Claim now through gRPC  ( In relation to issue 134 from snet-cli) 

anandsnet:
The below are now supported in Daemon through gRPC
 “list-unclaimed” - Get list of unclaimed payments , Get list of unclaimed payments, we do this by getting the list of channels in progress which have some amount to be claimed.

“start-claim” : Daemon will increase the nonce of the channel and send back the last payment with old nonce. 

“list-in-progress” - Get list of payments which are in progress. ( claims in progress)

Please refer to 
https://github.com/singnet/snet-cli/issues/134

No write operation on block chains are done by Daemon (will be take care of by the snet client )
Finish on the claim should be called only after the payment is successfully claimed and block chain is updated accordingly.
One way to determine this is by checking the nonce in the block chain with the nonce in the payment,
for a given channel if the block chain nonce is greater than that of the nonce from etcd storage => that the claim is already done in block claim and the Finish method is called on the claim.

Daemon will verify the below
1) mpe_address passed is correct
2) actual block_number is not very different (+-5 blocks) from the current_block_number from the signature
3) Verify that message was signed by the service provider (“payment_address” in metadata should match to the signer).


anandsnet:
@astroseger @raamb @Dinesh2521 @vsbogd 
Request your review on the this.

anandrgit:
Please note , will raise a separate pull request updating the README.md files on Daemon 
and removing references to channel claim command)

astroseger:
@anandrgit  
There is a rather strange bug. 
First time when I call GetListInProgress it return list of payments with empty signature field. But when I call it second time it return correct signature and everything works perfectly.

astroseger:
@anandrgit
Actually the problem is following StartClaim always sends empty signature field, so I can get signature only with GetListInProgress. But daemon should return full payment (including signature) as response to StartClaim.

anandrgit:
@astroseger , added the signature as part of the response being returned from claim

astroseger:
@anandrgit 
there is one more issue with StartClaim. It doesn't return correct signed amount (you do not set signed_amount at all in beginClaimOnChannel).

astroseger:
It works now. I "approve" this PR from the point of view of functionality. But proper review should be made by someone who know go.

escrow/control_service.go
```diff
@@ -0,0 +1,315 @@
...
+	}
+	differenceInBlockNumber := latestBlockNumber.Cmp(blockNumberPassed)
+	if -5 <= differenceInBlockNumber || differenceInBlockNumber <= 5 {
+		return nil
+	}
+	return fmt.Errorf("difference between the latest block chain number and the block number passed is %v ", blockNumberPassed)
+}
+
+//Get the current block number from on chain
+func currentBlock() (*big.Int, error) {
```
vsbogd: I think this method can be replaced by `Processor.CurrentBlock()` see https://github.com/singnet/snet-daemon/blob/05b4021d74e0516191289d3b95f68d1f94132385/blockchain/blockchain.go#L114

vsbogd:
Hi @anandsnet, sorry for a late review, code looks good and clean for me. The only comment I have is [about currentBlock() code duplication](https://github.com/singnet/snet-daemon/pull/184#pullrequestreview-192104520) I don't think it is blocker but may be you can fix it later. Thanks.

escrow/control_service.go
anandrgit: Thanks @vsbogd , was looking at making the  CurrentBlock  a Function rather than a Method ? this way we dont need an instance of *Processor.

anandrgit: thanks @vsbogd , will track this as a separately [#189](https://github.com/singnet/snet-daemon/issues/189)

Issue #173, opened 2018-12-12
Provide an interface for unlocking channels after application crash.


vsbogd:
Daemon may fail to cleanup state correctly in case it crashes. If there are any transactions were in progress at this moment they will stay locked, see https://github.com/singnet/snet-daemon/issues/172. So we probably need some management interface to deal with, see https://github.com/singnet/snet-daemon/issues/172#issuecomment-446765579

Before we have it in place we need to document manual resolution procedure.

vsbogd:
> should we unlock the channel when the daemon is starting up?

We cannot unlock channels on daemon start as there are may be more then one instance of daemon, in such case unlock logic will be broken.

Ideally I would say this functionality should be a part of "Operator UI" described in issue #29. What kind of UI should it be it is another question which is better to discuss in #29. I had experience of using Java JMX to add web interfaces to the service processes to review and manipulate by process internal state directly via HTTP. It is very powerful tools which can help to solve many complex things quickly.

Taking into account time restrictions for Beta, I would add additional command line interface, because it is straightforward thing. And then move this functionality to the "Operator UI" when issue #29 is implemented.

Dinesh2521:
@vsbogd 
need help with few clarifications
1. Can we assume that the user already knows the channel ID to Unlock? or we should give an option to show the locked channel IDs?
2. How many locked channels we will have? only one or it could be more?

planning to add new command group for channel related commands
example command to unlock a channel would be : **snetd channel --unlock <channelID>**
we can keep add other channel related commands as need

vsbogd:
> Can we assume that the user already knows the channel ID to Unlock? or we should give an option to show the locked channel IDs?

Yes, we can assume user should already know channel ID to unlock, because there are many ID can be locked simultaneously and it manual investigation is required to find out improperly locked id. One need to check that there are no operation on this channel is in progress using daemon logs for instance (see issue #172).

> How many locked channels we will have? only one or it could be more?

Is your question about do we need to support unlock for many channels in one command? The answer depends on daemon usage profile. I would say it should be a rare situation so for now we can have command to unlock one channel and use bash script to unlock many channels if it is needed.

Dinesh2521:
We will be removing the key from etcd to unlock the channels.
unlock command would look like 
`$ sned channel --unlock {channel ID}`

raamb:
PR #187 resolves this issue.

PR #187, opened 2019-01-12
Provide an interface for unlocking channels after application crash (issue #173)

Dinesh2521:
Daemon may fail to cleanup state correctly in case it crashes. If there are any transactions were in progress at this moment they will stay locked.  
We can unlock the channel using unlock command from daemon
`$ snetd channel --unlock {channel ID}`

example, if the channel id is 42 then the command would be
`$ snetd channel --unlock 42`

@raamb @anandrgit @vsbogd  can you please review and provide your comments/suggestions

### DIFF
--- a/escrow/lock.go
+++ b/escrow/lock.go
@@ -19,12 +19,12 @@ type Locker interface {
 // NewEtcdLocker returns new lock which is based on etcd storage.
 func NewEtcdLocker(storage AtomicStorage) Locker {
 	return &etcdLocker{
-		storage: NewPrefixedStorage(storage),
+		storage: NewLockerStorage(storage),
 	}
 }
 
 // returns new prefixed storage
-func NewPrefixedStorage(storage AtomicStorage) *PrefixedAtomicStorage {
+func NewLockerStorage(storage AtomicStorage) *PrefixedAtomicStorage {
 	return &PrefixedAtomicStorage{
 		delegate:  storage,
 		keyPrefix: "/payment-channel/lock",
--- a/escrow/lock.go
+++ b/escrow/lock.go
@@ -19,10 +19,15 @@ type Locker interface {
 // NewEtcdLocker returns new lock which is based on etcd storage.
 func NewEtcdLocker(storage AtomicStorage) Locker {
 	return &etcdLocker{
-		storage: &PrefixedAtomicStorage{
-			delegate:  storage,
-			keyPrefix: "/payment-channel/lock",
-		},
+		storage: NewPrefixedStorage(storage),
+	}
+}
+
+// returns new prefixed storage
+func NewPrefixedStorage(storage AtomicStorage) *PrefixedAtomicStorage {
+	return &PrefixedAtomicStorage{
+		delegate:  storage,
+		keyPrefix: "/payment-channel/lock",
 	}
 }
 
--- a/snetd/cmd/channel.go
+++ b/snetd/cmd/channel.go
@@ -73,7 +73,7 @@ func (command *channelCommand) unlockChannel() (err error) {
 	// verify whether the key exists or not
 	_, ok, _ := command.etcdclient.Get(channelKey)
 	if !ok {
-		fmt.Println("Error: Channel is not found by key:", key.String())
+		fmt.Println("Error: Channel is not found by key:", channelKey)
 	}
 	// if exists, delete the key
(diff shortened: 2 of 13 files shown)


snetd/cmd/channel.go
```diff
@@ -0,0 +1,80 @@
...
+	paymentChannelId *big.Int
+}
+
+// initializes and returns the new channel command object
+func newChannelCommand(cmd *cobra.Command, args []string, components *Components) (command Command, err error) {
+	channelId, err := getPaymentChannelId(cmd)
+	if err != nil {
+		return
+	}
+	if err != nil {
```
raamb: The if check is repeating, merge issue maybe

snetd/cmd/flags.go
```diff
@@ -53,6 +53,8 @@ const (
 	ClaimPaymentIdFlag = "payment-id"
 	ClaimSendBackFlag  = "send-back"
 	ClaimTimeoutFlag   = "timeout"
+
```
raamb: Remove newline here and  at line 83

Dinesh2521: fixed

snetd/cmd/channel.go
Dinesh2521: removed the duplicate check

snetd/cmd/channel.go
```diff
@@ -0,0 +1,76 @@
...
+		return fmt.Errorf("--unlock channel-id must be set")
+	}
+	if command.paymentChannelId != nil {
+		return command.unlockChannel()
+	}
+	return
+}
+
+// unlocks the channel with a given channel ID
+func (command *channelCommand) unlockChannel() (err error) {
```
vsbogd: I would re-use same `PrefixedAtomicStorage` which is used in `EtcdLocker` see https://github.com/singnet/snet-daemon/blob/05b4021d74e0516191289d3b95f68d1f94132385/escrow/lock.go#L20-L27
otherwise you have a risk that someone changes the locker code and will not change yours code and will break the functionality. 

I would do the following: 
- introduce new component etcdLockerStorage
- initialize it as in etcdLocker constructor: https://github.com/singnet/snet-daemon/blob/05b4021d74e0516191289d3b95f68d1f94132385/escrow/lock.go#L22-L25
- use this component in `NewEtcdLocker` and in `channelComand` (instead of etcdClient).
- use `PrefixedAtomicStorage.Delete()` method to remove the key

Dinesh2521: made the changes as discussed

snetd/cmd/components.go
```diff
@@ -141,6 +142,14 @@ func (components *Components) EtcdClient() *etcddb.EtcdClient {
 	return components.etcdClient
 }
 
+func (components *Components) PrefixedAtomicStorage() *escrow.PrefixedAtomicStorage {
```
vsbogd: Let's rename it to something less abstract for instance `EtcdLockerStorage` or `LockerStorage`. `PrefixedAtomicStorage` type is used few times in code for different storages.

escrow/lock.go
```diff
@@ -19,10 +19,15 @@ type Locker interface {
...
-		storage: &PrefixedAtomicStorage{
-			delegate:  storage,
-			keyPrefix: "/payment-channel/lock",
-		},
+		storage: NewPrefixedStorage(storage),
+	}
+}
+
+// returns new prefixed storage
+func NewPrefixedStorage(storage AtomicStorage) *PrefixedAtomicStorage {
```
vsbogd: Same thing as in https://github.com/singnet/snet-daemon/pull/187/files#r248655810 Let's rename it to something less abstract for instance `NewEtcdLockerStorage` or `NewLockerStorage`. `PrefixedAtomicStorage` type is used few times in code for different storages.

snetd/cmd/channel.go
vsbogd: @Dinesh2521, thank you looks perfect to me, just rename the component and constructor to something less abstract, thanks.

snetd/cmd/components.go
Dinesh2521: LockerStorage would be ideal i think. And changed the code accordingly.

escrow/lock.go
Dinesh2521: LockerStorage would be ideal i think. And changed the code accordingly.

Issue #175, opened 2018-12-18
Take Registry address from platform-contracts by default.


astroseger:
Daemon should take Registry address from platform-contracts by defaults.

Registry address is already here: resources/blockchain/node_modules/singularitynet-platform-contracts/networks. This file contains dictionary: netowrkID -> "contract addresses" 
1. At compilation time: we should take networks and store than as a constant
2. At execution time we should take networkId from RPC endpoint, ans use stored contract addresses by default.

vsbogd:
related to #142

PR #190, opened 2019-01-17
Issue #175 : Auto determine the address of the Registry based on the Network Selected

anandrgit:
Based on the network selected ( kovan /ropsten/local/main)  , we identify the network ID it is mapped to , using this network ID , the registry address will be read from the networks/Registry.json file .
With this change we get away from manually setting the registry and the ethereum end point in the daemon configuration.( snetd.config.json)

Take Registry address from platform-contracts by default. beta
#175

### DIFF
--- a/config/blockchain_network_config.go
+++ b/config/blockchain_network_config.go
@@ -4,6 +4,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"github.com/pkg/errors"
+	log "github.com/sirupsen/logrus"
 	"io/ioutil"
 )
 
@@ -54,7 +55,8 @@ func determineNetworkSelected(data []byte) {
 	networkName := GetString(BlockChainNetworkSelected)
 	err := json.Unmarshal(data, &dynamicBinding)
 	if err != nil {
-		panic(err)
+		panic(fmt.Sprintf("cannot parse the JSON configuation file %v for the network %v  to determine the network selected: %v",
+			BlockChainNetworkFileName, GetString(BlockChainNetworkSelected), err))
 	}
 	networkSelected = &NetworkSelected{
 		//Get the Network Name selected in config ( snetd.config.json) , Based on this retrieve the Registry address ,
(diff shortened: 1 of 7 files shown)


config/blockchain_network_config.go
```diff
@@ -0,0 +1,140 @@
...
+    "ethereum_json_rpc_endpoint":"https://mainnet.infura.io",
+    "network_id":"1"
+  }
+}`
+	BlockChainNetworkFileName  = "blockchain_network_config.json"
+	EthereumJsonRpcEndpointKey = "ethereum_json_rpc_endpoint"
+	NetworkId                  = "network_id"
+	RegistryAddressKey         = "registry_address_key"
+	//Read this from a config file path todo
+	RegistryJsonFileName = "resources/blockchain/node_modules/singularitynet-platform-contracts/networks/Registry.json"
```
vsbogd: Not sure why this file would be readable? It is a long filename, file as far as I understand should be inside NPM package. How will this path be available from daemon working directory?

config/blockchain_network_config.go
```diff
@@ -0,0 +1,140 @@
...
+)
+
+var networkSelected *NetworkSelected
+
+func determineNetworkSelected(data []byte) {
+	dynamicBinding := map[string]interface{}{}
+	networkName := GetString(BlockChainNetworkSelected)
+	err := json.Unmarshal(data, &dynamicBinding)
+	if err != nil {
+		panic(fmt.Sprintf("cannot parse the JSON configuation file %v for the network %v  to determine the network selected: %v",
```
vsbogd: I would replace `panic` calls here and below by returning `err` to the caller and further and at the end return it from `setAndValidateBlockChainNetworkDetails` and from `Validate` function in `config.go`. Thus all validation errors will be handled by similar way but I don't insist on it.

config/blockchain_network_config.go
anandrgit: you are right , will fix this post Beta , closing this PR for now

config/blockchain_network_config.go
anandrgit: Will do

Issue #172, opened 2018-12-11
"another transaction on channel: {ID: 20} is in progress"


arturgontijo:
Hi guys, I have a channel to call `basic-service` that I had published a time ago.
Its state:
```
$ snet client get_channel_state 20 54.203.198.53:7002
current_nonce                  = 0
current_signed_amount_in_cogs  = 5
current_unspent_amount_in_cogs = 106
```

But when I call the service using it, I'm getting this error:
```
$ snet client call 20 0.00000001 54.203.198.53:7002 mul '{"a":15,"b":78}'
unspent_amount_in_cogs before call (None means that we cannot get it now):106
Error: <_Rendezvous of RPC that terminated with:
	status = StatusCode.FAILED_PRECONDITION
	details = "another transaction on channel: {ID: 20} is in progress"
	debug_error_string = "{"created":"@1544558848.135429276","description":"Error received from peer","file":"src/core/lib/surface/call.cc","file_line":1036,"grpc_message":"another transaction on channel: {ID: 20} is in progress","grpc_status":9}"
>
If you want to see full Traceback then run:
snet --print-traceback [parameters]
```

Then I opened a new channel and the call function worked fine:
```
$ snet client call 181 0.00000001 54.203.198.53:7002 mul '{"a":15,"b":78}'
unspent_amount_in_cogs before call (None means that we cannot get it now):1
value: 1170.0
```

What I have to do to "terminate" the pending transaction from channel 20?

vsbogd:
@arturgontijo , thank you for reporting this. The reason is your etcd storage contains a key which locks transactions for the channel. Daemon aquired lock but did not release it. It might happen easily with old versions of daemon, it should not be the case since https://github.com/singnet/snet-daemon/pull/154 is merged. 

If you don't need the data you currently have in your etcd storage you can remove `storage-data-dir-1.etcd` folder it will eliminate the error. If you need the data then you can fix it by removing key `'/payment-channel/lock/{ID: 20}` from etcd. Right now it should be possible to use `etcdctl` tool for doing this. 

But I think we need an interface to manage it by simpler way. Something like "Operator UI" mentioned at https://github.com/singnet/snet-daemon/issues/29 .

vsbogd:
Raised https://github.com/singnet/snet-daemon/issues/173

arturgontijo:
> It might happen easily with old versions of daemon, it should not be the case since #154 is merged.

[...]

PR #191, opened 2019-01-17
Channel lock Issue #172 ( income % does not equal to price % )

anandrgit:
Once you hit the error "income % does not equal to price % ", the channel will get locked as the rollback wasn't happening.

This change is to fix it , details of the fix are below
1) define the defer function call well before the first return on an error ( else the defer function doesnt get called)
Since the Rollback is handled at centralized place () intercept ,
2) make sure the transactions on the method called are is returned even if there is an error ( else the rollback will not happen and the lock is set on the channel)
Also related to singnet#172
singnet#172

### DIFF
--- a/escrow/payment_handler.go
+++ b/escrow/payment_handler.go
@@ -60,14 +60,14 @@ func (h *paymentChannelPaymentHandler) Payment(context *handler.GrpcStreamContex
 
 	transaction, e := h.service.StartPaymentTransaction(internalPayment)
 	if e != nil {
-		return nil, paymentErrorToGrpcError(e)
+		return transaction, paymentErrorToGrpcError(e)
 	}
 
 	income := big.NewInt(0)
 	income.Sub(internalPayment.Amount, transaction.Channel().AuthorizedAmount)
 	e = h.incomeValidator.Validate(&IncomeData{Income: income, GrpcContext: context})
 	if e != nil {
-		return nil, paymentErrorToGrpcError(e)
+		return transaction, paymentErrorToGrpcError(e)
 	}
 
 	return transaction, nil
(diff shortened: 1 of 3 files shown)


escrow/payment_handler.go
```diff
@@ -60,14 +60,14 @@ func (h *paymentChannelPaymentHandler) Payment(context *handler.GrpcStreamContex
 
 	transaction, e := h.service.StartPaymentTransaction(internalPayment)
 	if e != nil {
-		return nil, paymentErrorToGrpcError(e)
+		return transaction, paymentErrorToGrpcError(e)
```
raamb: Can transaction be a non-null object in this case? Since the error is from the StartPaymentTransaction method I assume that transaction would be nil if an error has occurred

anandrgit:
Closing this PR, Will refactor this more and raise a fresh PR

vsbogd:
@anandrgit, thanks for catching it. Regarding fix, I would instead fix `Payment` method to rollback transaction if price validation is not passed. 

```golang
func (h *paymentChannelPaymentHandler) Payment(context *handler.GrpcStreamContext) (payment handler.Payment, err *handler.GrpcError) {
	internalPayment, err := h.getPaymentFromContext(context)
	if err != nil {
		return
	}

	transaction, e := h.service.StartPaymentTransaction(internalPayment)
	if e != nil {
		return nil, paymentErrorToGrpcError(e)
	}
        // new defer statement
	defer func() {
		if err != nil {
			transaction.Rollback();
		}
	}

	income := big.NewInt(0)
	income.Sub(internalPayment.Amount, transaction.Channel().AuthorizedAmount)
	e = h.incomeValidator.Validate(&IncomeData{Income: income, GrpcContext: context})
	if e != nil {
                // assign err to make defer work properly
		err = paymentErrorToGrpcError(e)
		return
	}

	return transaction, nil
}
```

Another way is to move validation into `StartTransaction` method.

anandrgit:
Thanks @vsbogd 
We realised this too just sometime back :)  and hence  closed this Pull Request , and raised a new one 
https://github.com/singnet/snet-daemon/pull/193

PR #198, opened 2019-01-22
Point to the latest version of the contract 0.2.6

anandrgitnirman:
Point to the latest version of the contract 0.2.6


astroseger:
In any case. Even we don't use token contract address in Daemon we still need it in snet-cli. So for sake of consistency it is better to use the same new version of token contract in the daemon and in the snet-cli.

Issue #199, opened 2019-01-23
daemon behind the NAT


astroseger:
It seems in the current version we cannot configure the daemon to work behind NAT.
In the case of working behind NAT we should set DAEMON_END_POINT to an internal endpoint. But  daemon search for group_id using DAEMON_END_POINT and it will not find it, because in metadata we have an external endpoint..

anandrgitnirman:
Thanks @astroseger 
on Solution 1, I didn't understand this bit "GROUP_NAME can be specified only if DAEMON_END_POINT is not in metadata",  

Could we consider using Group_name as a mandatory field itself in Daemon? and get away from Daemon End point config in Daemon ? , we will then just have the Group Name and the port number as mandatory configs in daemon  ( this will be used by the daemon to start) 

@raamb  too for notification

astroseger:
Correction: In the current version we have a problem only if external and internal ports are different. If ports are the same we should set DAEMON_END_POINT to the external endpoint and everything will work, because daemon will listen everything on this specific port.

Concerning "GROUP_NAME can be specified only if DAEMON_END_POINT is not in metadata" in Solution 1. It is only to prevent errors. If DAEMON_END_POINT is in metadata then GROUP_NAME shoudn't be specified. Or at least we should check that it is a correct GROUP_NAME (that DAEMON_END_POINT belongs to this group).

I'm ok with GROUP_NAME + PORT solution. But it will make old config invalid... It might be not ok now.

There is another solution... We could have an optional parameter DAEMON_PORT... If daemon port is not specified we get it from DAEMON_END_POINT.

ksridharbabuus:
@astroseger The standard approach will be providing Daemon_port Parameter rather than combing the Port with Group_Name. 
I have one Question... Is all the Daemon should have the same port in the Group? 
We also had various other discussions in the past like a Load Balance / Routing Component and so on... Better to consider all possible scenarios while coming up with the solution.

astroseger:
@ksridharbabuus 
I think that we shouldn't require that all daemons in the same payment group have the same port.

In any case, we should allow the possibility to have different external and internal ports (in case of NAT). It means that in config we should be able to set (directly or indirectly) two independent parameters:
- port. This port can be different from the port in metadata (in case then internal and external ports are different). 
- payment group.

ferrouswheel:
[...]

PR #202, opened 2019-01-28
Issue #199  Behind NAT support 

anandrgitnirman:
Issue #199 

Get away from the Daemon End Point
Introduce group name to help with payments
Define the port number that the Daemon should listen to .

The issue with having a daemon end point was that it was forcing the endpoint defined to be exactly the same as what has been defined in the metadata ( this may not always be true as the daemon could run in a private network behind the NAT)
What has been defined in the meta data is for public access, how daemon runs in its private network or container should be agnostic to this public end point

### DIFF
--- a/README.md
+++ b/README.md
@@ -139,11 +139,11 @@ configuration properties can be set using configuration file.
 These properties you should usually change before starting daemon for the first
 time.
 
-* **daemon_endpoint** (optional; default: `"127.0.0.1:8080"`) - 
-network interface and port which daemon listens to. This parameter should be
-absolutely equal to the corresponding endpoint in the [service configuration
-metadata][service-configuration-metadata]. URI format is recommended:
-http://<host>:<port>.
+* **daemon_group_name** (required ,default: `"default_group"`) - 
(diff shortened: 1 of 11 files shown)


anandrgitnirman:
@vsbogd @raamb @astroseger 
Request your review on this change

astroseger:
I propose to make ```daemon_group_name``` optional (with default = "default_group").

anandrgitnirman:
hi @astroseger , corrected this in the Readme , code wise we have the daemon group as default_group

ferrouswheel:
Nice!

One thing is that this seems renames the daemon end point config key. While, I can see a reason to do it... this will require everyone to change their config and for us to update our documentation. Created an issue on the dev-portal for this.

astroseger:
It seems the only one problem is that now ```http://``` prefix is now allowed for daemon_end_point

ferrouswheel:
@astroseger Ah yeah. I just saw the change to the documentation and assumed it had changed... but it's the same, the documentation was just being fixed. So ignore me haha :-)

PR #219, opened 2019-02-19
Selectively enable the SSL LetsEncrypt (ACME) http-01 challenge-response on port 80 

anandrgitnirman:
option to enable the SSL LetsEncrypt (ACME) http-01 challenge-response.
This was done to enable https on daemon using certbot by default.


anandrgitnirman:
obsolete

PR #223, opened 2019-02-23
Compiling .proto files for metrics service and escrow service in the …

vforvalerio87:
…installation script

### DIFF
--- a/resources/blockchain/package.json
+++ b/resources/blockchain/package.json
@@ -3,7 +3,7 @@
     "bip39": "2.5.0",
     "fs-extra": "5.0.0",
     "ganache-cli": "6.1.6",
-    "singularitynet-platform-contracts": "0.3.0" ,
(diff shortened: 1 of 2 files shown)


vsbogd:
@vforvalerio87, could you please remind why it is needed? 
Last line of the installation script: https://github.com/singnet/snet-daemon/blob/baab198001ae2ef3142ef8b7f7e6f24d6b4c3e6f/scripts/install#L19  plus `go:generate` definitions in corresponding files should be enough: https://github.com/singnet/snet-daemon/blob/baab198001ae2ef3142ef8b7f7e6f24d6b4c3e6f/escrow/state_service.go#L1

vforvalerio87:
Strange, that didn't work for me. Maybe it's OS-related? I always had to compile those manually. Let me run a couple of tests

PR #229, opened 2019-03-19
Determine the current version of Daemon through cli commands

anandrgitnirman:
using ldflags , latest tagged version , sha1 revision and the buld time are set at the compile time , 
these versioning details can  now easily  be sent back on the response from Daemon

### DIFF
--- a/scripts/build
+++ b/scripts/build
@@ -6,5 +6,8 @@ PARENT_PATH=$(dirname $(cd $(dirname $0); pwd -P))
 
 pushd $PARENT_PATH
 mkdir -p build
-GOOS=$1 GOARCH=$2 go build -o build/snetd-$1-$2 snetd/main.go
+now=$(date +'%Y-%m-%d_%T')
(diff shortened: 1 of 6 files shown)


scripts/build
```diff
@@ -6,5 +6,8 @@ PARENT_PATH=$(dirname $(cd $(dirname $0); pwd -P))
 
 pushd $PARENT_PATH
 mkdir -p build
-GOOS=$1 GOARCH=$2 go build -o build/snetd-$1-$2 snetd/main.go
+now=$(date +'%Y-%m-%d_%T')
```
raamb: How is the version incremented?

anandrgitnirman: version tags will be added before this script is run ( @vinthedark  and I discussed this as part of the build)

raamb:
Please add a description to the PR on what changes have been made. Will be easy to understand going forward

scripts/build
vsbogd: @anandrgitnirman , please update documentation to mirror new parameter of the build script

PR #230, opened 2019-03-20
Support cross compilation with xgo

### LINKED ISSUE #215: Unable to  build Mac Binary of Daemon from Ubuntu

### LINKED ISSUE #227: ARM architecture support
I'm investigating ARM support, so that snetd can be run on low-powered devices like the Raspberry Pi and be used in IoT applications.

This issue is an attempt to log my progress.

Initially I'm trying to cross compile with the following:

```
env GOOS=linux GOARCH=arm GOARM=6 go build -o build/snetd-linux-arm6 snetd/main.go
```

The first issue is:
```
# github.com/singnet/snet-daemon/ratelimit
ratelimit/rateLimit.go:15:13: constant 9223372036854775807 overflows int
```
That can be fixed by using `math.MaxInt32` instead of `math.MaxInt64`.

The next issue is that ethereum-go has to do some C compilation and [requires xgo for cross-compilation to work properly](https://github.com/ethereum/go-ethereum/wiki/Cross-compiling-Ethereum).

```
build constraints exclude all Go files in /home/joel/go/src/github.com/singnet/snet-daemon/vendor/github.com/ethereum/go-ethereum/crypto/secp256k1
```

While cross compilation would make development faster and more convenient, tomorrow I'll try compiling directly on the Pi.

ferrouswheel:
I've added a script using xgo, to allow us to build for arm6 (raspberry pi). Fixes #227

Side benefit of this is approach is that I was able to build for almost every platform Go supports (so also fixes #215).

However, I've constrained the default behaviour to only build for the major platforms.

Also changes some Int64 constants to fit in Int32 so compilation on arm works.

### DIFF
--- a/ratelimit/rateLimit.go
+++ b/ratelimit/rateLimit.go
@@ -12,7 +12,7 @@ func NewRateLimiter() rate.Limiter {
 	//By Default set the maximum value possible for the Burst Size ( assuming rate was defined ,but burst was not defined)
 	burstSize := config.GetInt(config.BurstSize)
 	if burstSize == 0 {
-		burstSize = math.MaxInt64
+		burstSize = math.MaxInt32
 	}
 	limiter := rate.NewLimiter(getLimit(), burstSize)
 	return *limiter
--- a/scripts/watcher_fsevents_cgo.go.patch
+++ b/scripts/watcher_fsevents_cgo.go.patch
@@ -1,0 +1,22 @@
+--- /home/joel/Downloads/watcher_fsevents_cgo.go	2019-03-20 13:14:12.643748593 +1300
++++ watcher_fsevents_cgo.go	2019-03-20 12:49:33.489170174 +1300
+@@ -48,7 +48,7 @@
+ // started and is ready via the wg. It also serves purpose of a dummy source,
+ // thanks to it the runloop does not return as it also has at least one source
+ // registered.
+-var source = C.CFRunLoopSourceCreate(nil, 0, &C.CFRunLoopSourceContext{
++var source = C.CFRunLoopSourceCreate(C.kCFAllocatorDefault, 0, &C.CFRunLoopSourceContext{
+ 	perform: (C.CFRunLoopPerformCallBack)(C.gosource),
+ })
+ 
+@@ -162,8 +162,8 @@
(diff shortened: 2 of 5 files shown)


ferrouswheel:
Not sure whether to also swap circleci to using xgo or if there are packaging scripts that should use this for releases, let me know if you want me to work on that too. cc @anandrgitnirman

ratelimit/rateLimit.go
```diff
@@ -12,7 +12,7 @@ func NewRateLimiter() rate.Limiter {
 	//By Default set the maximum value possible for the Burst Size ( assuming rate was defined ,but burst was not defined)
 	burstSize := config.GetInt(config.BurstSize)
 	if burstSize == 0 {
-		burstSize = math.MaxInt64
+		burstSize = math.MaxInt32
```
raamb: Was this change required for the cross compilation?

scripts/watcher_fsevents_cgo.go.patch
```diff
@@ -0,0 +1,22 @@
+--- /home/joel/Downloads/watcher_fsevents_cgo.go	2019-03-20 13:14:12.643748593 +1300
```
raamb: Is this file related to cross compilation?

ratelimit/rateLimit.go
ferrouswheel: Yes, Int is 32bit on Arm6

scripts/watcher_fsevents_cgo.go.patch
ferrouswheel: Yes, at least until we upgrade our deps and Go. This patch is required for compiling for OSX and gets automatically applied by the build script.

ratelimit/rateLimit.go
raamb: Thanks

scripts/watcher_fsevents_cgo.go.patch
raamb: Thanks, that clarifies

raamb:
@vinthedark once this is merged in, lets updated out build scripts to generate releases with binaries for all platforms

PR #231, opened 2019-03-20
Add config check to validate daemon and passthrough endpoints

### LINKED ISSUE #217: Check that the passthrough endpoint is different to the daemon endpoint
It is currently possible to make the passthrough endpoint the same as the daemon endpoint. This results in the daemon calling itself and reporting that the channel is locked, which was the cause #210.

To avoid this user error, it would be nice to for the daemon to validate that the two endpoints are different.

An example of the offending `snetd.config`:

```
{
   "DAEMON_END_POINT": "0.0.0.0:6206",
   "ETHEREUM_JSON_RPC_ENDPOINT": "https://kovan.infura.io",
   "IPFS_END_POINT": "http://ipfs.singularitynet.io:80",
   "REGISTRY_ADDRESS_KEY": "0xe331bf20044a5b24c1a744abc90c1fd711d2c08d",
   "PASSTHROUGH_ENABLED": true,
   "PASSTHROUGH_ENDPOINT": "http://127.0.0.1:6206",
   "ORGANIZATION_ID": "snet",
   "SERVICE_ID": "semantic-segmentation",
   "LOG": {
       "LEVEL": "debug",
       "OUTPUT": {
            "TYPE": "stdout"
           }
   }
}
```

As shown, it won't be enough to just do an exact match, since 0.0.0.0 will bind to all interfaces, and the passthrough endpoint has the url scheme prefix.

ferrouswheel:
Mainly to check they are not the same and fix #217

Doesn't catch any ipv6 stuff, or whether a domain resolves to the same IP, but should catch the bad config I was using that had the daemon calling itself, and that confused me for a couple of days! :sweat_smile:

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"math/big"
 	"net/url"
+	"net"
 	"regexp"
 	"sort"
 	"strings"
@@ -175,6 +176,13 @@ func Validate() error {
 		!IsValidUrl(vip.GetString(MonitoringServiceEndpoint)) {
 		return errors.New("service endpoint must be a valid URL")
 	}
+	passEndpoint := vip.GetString(PassthroughEndpointKey)
+	daemonEndpoint := vip.GetString(DaemonEndPoint)
+	var err error
+	err = ValidateEndpoints(daemonEndpoint, passEndpoint)
+	if err != nil {
+		return err
+	}
 
 	// Validate metrics URL and set state
 	return nil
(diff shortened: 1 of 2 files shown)


config/config.go
```diff
@@ -268,3 +276,25 @@ func ValidateEmail(email string) bool {
...
+	if err != nil {
+		return errors.New("couldn't split host:post of daemon endpoint")
+	}
+
+	if daemonHost == passthroughURL.Hostname() && daemonPort == passthroughURL.Port() {
+		return errors.New("passthrough endpoint can't be the same as daemon endpoint!")
+	}
+
+	if ((daemonPort == passthroughURL.Port()) &&
+	    (daemonHost == "0.0.0.0") &&
```
anandrgitnirman: we will have a problem when say daemonHost="localhost"  and daemonHost != passThroghHost no ?

ferrouswheel: I don't understand what problem you are thinking of. Could you give an example of the two endpoint strings, and what you'd expect to have happen?

I have added tests so it's clear what values should or shouldn't work. Should be easy to add a new one if you want to confirm your example.

anandrgitnirman: Apologies should  have been more clear ,  the example below will also be a case where daemon is calling itself  no? 
  "passthrough_endpoint": "http://0.0.0.0:7003",
  "daemon_end_point": "localhost:7003",

ferrouswheel:
0.0.0.0 is invalid for passthrough, which needs to be a specific address.

0.0.0.0 isn't a real address, but for servers it means "bind to all local
interfaces"

config/config.go
anandrgitnirman: I am good, @Dinesh2521  if you dont have any further comments , we cal close this PR today itself

Issue #192, opened 2019-01-18
Remove `claim` command from daemon and documentation


vsbogd:
As `snet-cli` should be used to claim payments, old `snet-daemon claim` command should be removed and all related blockchain code, see issue https://github.com/singnet/snet-daemon/issues/106 for example.

PR #234, opened 2019-03-22
#192 : Remove claim  command in Daemon 

anandrgitnirman:
With all claims now managed by snet-cli , Obsolete Code to claim  in Daemon is being removed 
https://github.com/singnet/snet-daemon/issues/192


metrics/heartbeat.py
```diff
@@ -0,0 +1,31 @@
+
```
raamb: Why is the heartbeat file appearing in the claim PR

vsbogd:
Looks good to me, except `heartbeat.py` file

heartbeat.py
```diff
@@ -0,0 +1,31 @@
+
```
vsbogd: It seems you moved it instead of removing

anandrgitnirman: my bad :) , removed it now for sure , also removed references to private keys and mnemonic

Issue #195, opened 2019-01-21
Streaming Requests and make the max message size a config key. 


anandrgit:
Encourage the use of streaming requests and make the max message size a config key. 
We shouldn't require everyone to accept 4GB messages.

assess  whether increasing the max message size impacts memory use?

If public APIs can be sent 4GB messages, and those messages are pre-allocated before being verified as a correct user, then it'd be pretty easy for a malicious actor to take down SNet services.

anandrgitnirman:
All, 

We were able to test streaming on Daemon successfully using the proto below 

------------------------------------------------------------------------------------------------------------------------------
syntax = "proto3";

package proto;

service ds {

  rpc UnaryRPC(Message) returns (Message) {}

  rpc ClientStreamingRPC(stream Message) returns (Message) {}

  rpc ServerStreamingRPC(Message) returns (stream Message) {}

  rpc BiDirectionalStreamingRPC(stream Message) returns (stream Message) {}
}

  message Message {
    string message = 1;
  }

--------------------------------------------------------------------------------------------------------------------------------

An observation was that if the service definition returns an io.EOF , then this caused the Daemon to treat this as an Error 

_ func (s * DaemonStreamingServer) BiDirectionalStreamingRPC(stream proto.Ds_BiDirectionalStreamingRPCServer) (error) {

....
return io.EOF 

} _

anandrgitnirman:
Discussed with @raamb  With Streaming support in Daemon  ( we dont see a need to keep the max size configurable, thus avoiding upper limit or lower limit checks additionally)

raamb:
@ferrouswheel with support for streams coming in, not sure of the value of making the limit configurable, as you pointed out in your comment this can be a vector of attack. I would prefer to reduce this limit to an MB level. Let us know your thoughts.

ferrouswheel:
I think it may be worth allowing it to be configurable. If people have fixed size requests they want to handle, but don't want to change their server to support streaming, we should allow them to use it.

**However**, the default should be much much lower than 2 GB!

I think 4MB is the default for grpc, and keeping our default the same would be sensible.

PR #239, opened 2019-04-03
Issue : 195 - Keep the Max Message size configurable

anandrgitnirman:
Rather than forcing everyone to have a max message size of math.MaxInt32, make it configurable now 
#195


config/config.go
```diff
@@ -32,6 +32,7 @@ const (
 	IpfsEndPoint                   = "ipfs_end_point"
 	IpfsTimeout                    = "ipfs_timeout"
 	LogKey                         = "log"
+	MaxMessageSize                 = "max_message_size"
```
raamb: This is currently in bytes, lets make it in MB or GB and handle conversion. Also we need to enforce a lower and upper limit.

Issue #224, opened 2019-02-25
After changing group_id treasurer stops working


astroseger:
Way to reproduce the bug:
1. Service is running and receiving payments. 
2. Service provider update metadata and intentionally or by mistake change group_id. It should be noted that it is strongly not recommended, and in the last version of snet-cli it is not possible by default (however it still can be done with ```--force``` option). In order to change group_id your could simple ```metadata-init``` which will reinitialize group_id, and after you update metadata in Registry by ```snet service update-metadata``` (but in the last version you will need to use force here).
3. Restart daemon
4. open new payment channel and make few calls from the client side.
5. Run ```snet treasurer claim-all```, and it will fail with "Channel received belongs to another group of replicas,...." error from the daemon...

The possible cause of the problem is following: then daemon execute treasurer related commands it sometime verify that group_id of the channel corresponds to the current group_id and sometimes not...:
- ```GetListUnclaimed``` return list of all payment channels (including channels with old group_id)
- ```StartClaim``` works only with channels with new group_id

I would suggest that in ```controle_service``` daemon should work with all channels with correct payment_address. The owner of ```payment_address``` (and we verify it with the signature) has right to do anything with his channels. So we shouldn't verify that group_id is still correct.. 
(even though the changing of group_id is not recommended)

anandrgitnirman:
hi @astroseger  with the latest change in snet-cli from what I understand it is not possible to modify an existing group_id right in metadata ?

astroseger:
Yes. service provider should never  do it. If your really need it you can force it by removing and adding service again. But all payment channels will became invalid, and we still have a bug in daemon https://github.com/singnet/snet-daemon/issues/224, so your daemon will also stop working.

astroseger:
@anandrgitnirman 
Yes it is impossible to do it via snet-cli. But still it is bug here. Service provider should be able to claim their funds even in case of changing group-id.

anandrgitnirman:
@vsbogd , @astroseger 

We have the  approaches to bypass the group ID check at the time of Claim
1) 
Introduce a new attribute _DisableGroupIdCheck bool_ under PaymentChannelKey   
https://github.com/singnet/snet-daemon/blob/e9ff650af297219ef2cafacf2ce88d7be6ddca29/escrow/payment_channel_api.go#L40-L42
then check for this flag at 
[...]

PR #240, opened 2019-04-03
Issue 224: Ensure Claim does not fail even when group ID changes 

anandrgitnirman:
https://github.com/singnet/snet-daemon/issues/224
Ensure Group ID check should only apply at the time of storing channel state in etcd for the first time
Moved the group ID check from BlockchainChannelReader to lockingPaymentChannelService.PaymentChannel()


anandrgitnirman:
@astroseger , tested this , works fine with claim now 

@vsbogd , request your review on this

escrow/escrow.go
```diff
@@ -56,6 +66,15 @@ func (h *lockingPaymentChannelService) PaymentChannel(key *PaymentChannelKey) (c
 	return MergeStorageAndBlockchainChannelState(storageChannel, blockchainChannel), true, nil
 }
 
+//Check if the channel belongs to the same group Id
+func (h *lockingPaymentChannelService) VerifyGroupId(configGroupID [32]byte ,blockChainGroupID  [32]byte ) error {
```
vsbogd: I would rename the method to `verifyGroupId` to make it private.

escrow/escrow_test.go
```diff
@@ -280,3 +281,41 @@ func (suite *PaymentChannelServiceSuite) TestStartClaim() {
 	assert.Equal(suite.T(), suite.payment(), claim.Payment())
 	assert.Equal(suite.T(), []*Payment{suite.payment()}, claims)
 }
+
+func (suite *PaymentChannelServiceSuite) TestVerifyGroupId() {
+
+
+	service := NewPaymentChannelService(
```
vsbogd: Test could be simplified and made more readable by copying `suite.service` and replacing only field required to replace:
```golang
service := suite.service()
service.(lockingPaymentChannelService).blockchainReader.replicaGroupID = func(channelID *big.Int) (*blockchain.MultiPartyEscrowChannel, bool, error) {
    return suite.mpeChannel(), true, nil
}
```

escrow/escrow_test.go
```diff
@@ -280,3 +281,41 @@ func (suite *PaymentChannelServiceSuite) TestStartClaim() {
 	assert.Equal(suite.T(), suite.payment(), claim.Payment())
 	assert.Equal(suite.T(), []*Payment{suite.payment()}, claims)
 }
+
+func (suite *PaymentChannelServiceSuite) TestVerifyGroupId() {
```
vsbogd: :+1:

escrow/escrow.go
```diff
@@ -45,8 +45,18 @@ func (h *lockingPaymentChannelService) PaymentChannel(key *PaymentChannelKey) (c
...
 	}
 
+
 	blockchainChannel, blockchainOk, err := h.blockchainReader.GetChannelStateFromBlockchain(key)
+
 	if !storageOk {
+		//Group ID check is only done for the first time , when the channel is added to storage from the block chain ,
+		//if the channel is already present in the storage the group ID check is skipped.
+	    if blockchainChannel != nil {
+			blockChainGroupID,err := h.blockchainReader.replicaGroupID()
```
vsbogd: As `replicaGroupId` is not used in blockchain reader anymore I would move it into `lockingPaymentChannelService` itself. But it will probably require fixing unit tests, so it makes sense to do separate PR for this.

escrow/escrow_test.go
anandrgitnirman: Thansk @vsbogd  , fixed this

escrow/escrow.go
anandrgitnirman: Will submit a separate PR by tomorrow , hopefully by then this is approved :)

PR #241, opened 2019-04-04
Issue : 195 -  make the max message size a config key.

anandrgitnirman:
Default value has been set to 4MB 
Any attempt to set the value  beyond 4GB will fail with an error

### DIFF
--- a/README.md
+++ b/README.md
@@ -212,8 +212,8 @@ metadata][service-configuration-metadata].
 * **log** (optional) - 
 see [logger configuration](./logger/README.md)
 
-* **max_message_size_in_mb** (optional; default: `4MB`) - 
-The default value set is to 4 MB, this is used to configure the max size in MB of the message received by the Daemon.
+* **max_message_size_in_mb** (optional; default: `4`) - 
+The default value set is to 4 (units are in MB ), this is used to configure the max size in MB of the message received by the Daemon.
 In case of Large messages , it is recommended to use streaming than setting a very high value on this configuration.
 It is not recommended to set the value more than 4GB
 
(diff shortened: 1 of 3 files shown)


snetd/cmd/serve.go
```diff
@@ -173,8 +172,8 @@ func (d daemon) start() {
 	}
 
 	if config.GetString(config.DaemonTypeKey) == "grpc" {
-		// set the maximum that the server can receive to 4GB. It is set to for 4GB because of issue here https://github.com/grpc/grpc-go/issues/1590
-		maxsizeOpt := grpc.MaxRecvMsgSize(math.MaxInt32)
+
+		maxsizeOpt := grpc.MaxRecvMsgSize(config.GetInt(config.MaxMessageSizeInMB) * 1024 * 1024)
```
ferrouswheel: This will overflow on 32 bit platforms like raspberry pi, whose maximum int value is value is 2,147,483,647

I think we should limit to this value. People should be using streaming before they get to the point of sending 2 GB messages, so it shouldn't be overly restricting to anyone.

README.md
```diff
@@ -212,6 +212,11 @@ metadata][service-configuration-metadata].
 * **log** (optional) - 
 see [logger configuration](./logger/README.md)
 
+* **max_message_size_in_mb** (optional; default: `4MB`) -
```
ferrouswheel: default should be just `4` - otherwise people will think they need to specify the `MB` suffix in the config value.

snetd/cmd/serve.go
anandrgitnirman: Enforced this check on the config , [validate](https://github.com/singnet/snet-daemon/pull/241/files#diff-3baf47c64847a8fb8aaa8cc2e088513bR190) function

PR #244, opened 2019-04-04
to make authorization in state_service the same as in control_service(fix for issue#228)

Dinesh2521:
this fix has 2 of the following changes
1. change the authorizaion logic in state service similar to the control service. as part of the message to be signed will contain the following fields now instead channel id before
- "__get_channel_state" text 
- channel ID
- current block number
we have added fall back to older message format as well, since it may break the interaction with CLI, if the developers are using older CLI versions. this backward compatibility can be removed in future releases.

2. moved some of the common authorization code to separate package, so that we can extend it to build the authentication service for upcoming changes like operator UI

added basic tests to verify the changes.  but we may have to review and add bit more detailed tests while developing it for operator UI

@vsbogd @astroseger can you please review the changes and let me know if incase of any changes/suggestion

### DIFF
--- a/escrow/state_service.proto
+++ b/escrow/state_service.proto
@@ -24,9 +24,11 @@ service PaymentChannelStateService {
 message ChannelStateRequest {
     // channel_id contains id of the channel which state is requested.
     bytes channel_id = 1;
+    //current block number (signature will be valid only for short time around this block number)
+    uint64 current_block = 2;
     // signature is a client signature of the message which contains
     // channel_id. It is used for client authorization.
-    bytes signature = 2;
+    bytes signature = 3;
 }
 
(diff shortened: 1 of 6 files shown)


escrow/state_service.proto
```diff
@@ -24,9 +24,11 @@ service PaymentChannelStateService {
 message ChannelStateRequest {
     // channel_id contains id of the channel which state is requested.
     bytes channel_id = 1;
+    //current block number (signature will be valid only for short time around this block number)
```
raamb: This will break backward compatibility. The new field should be assigned a new number

Dinesh2521:
to update it with new approach for stateless client logic

Issue #209, opened 2019-01-31
Make daemon handle signals (SIGINT, SIGTERM) and correctly cleanup the state


vsbogd:
At the moment if service provider kills daemon while it processes client request, then channel remains locked after daemon process is finished. The reason is that `Golang` doesn't call deferred functions automatically when process is finished due to signal received. 

To handle signals correctly in `Golang` it is required to:
- add signal channel and listen for the external signals
- make cleanup calls, close `gRPC` server
- gracefully shutdown all goroutines

The sketch of closing `gRPC` server on signal can be found at https://github.com/grpc/grpc-go/issues/1372

anandrgitnirman:
@vsbogd , just wanted to post the group that raam and I discussed this at length today , will send you a documentation on our approach in a day or two

anandrgitnirman:
@vsbogd @raamb 

****For a graceful Shutdown**** 

- *Add Signal channel and listen for external Signal* , 
This is already done on the Daemon code ``
https://github.com/singnet/snet-daemon/blob/v0.1.10/snetd/cmd/serve.go#L66-L70


- *To Stop new requests*
// GracefulStop stops the gRPC server gracefully. It stops the server from
// accepting new connections and RPCs and blocks until all the pending RPCs are
// finished.
So we just call the GracefulStop() and **not stop**


- *Existing requests*
->Ideal way would be to wait  for the existing requests to complete and handle the payment accordingly ?.Will need help here on how best to do this ( with support for streaming), and for later worry about asynch calls when we support that in Daemon 

We need to the below to gracefully shutdown go routines 

1. Signal goroutine to shutdown.
2. Wait for goroutine to shutdown.

->Shutdown all goroutines (we don't care if the request will be processed or not ?) , roll back the payment on any channels locked ( in progress) 

Check if any channels are locked because a request is currently being processed , then release the lock on these channels ALONE ?
We don't want to unlock channels locked for reasons other than the requests being processed , ? 
 
Other way is to use  wg := &sync.WaitGroup{}
call  wg.Wait() on Shutdown  
 `func cleanUp...(shutdown <-chan int, wg *sync.WaitGroup) {`

wg.Add(1) will be called after a channel has been locked 
wg.Done() will be called on defer function that 
//calls paymentHandler.Complete(payment) or //paymentHandler.CompleteAfterError(payment, e)
//In case of error while unlocking the channel , just log the error ?
 

[...]

PR #252, opened 2019-04-11
https://github.com/singnet/snet-daemon/issues/209

anandrgitnirman:
changes made inside a method with a pointer receiver is visible to the caller whereas this is not the case in value receiver
made start to have pointer receiver

### DIFF
--- a/snetd/cmd/serve.go
+++ b/snetd/cmd/serve.go
@@ -130,7 +130,7 @@ func newDaemon(components *Components) (daemon, error) {
 }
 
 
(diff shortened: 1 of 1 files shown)


snetd/cmd/serve.go
```diff
@@ -130,7 +130,7 @@ func newDaemon(components *Components) (daemon, error) {
 }
 
 
-func (d daemon) start() {
+func (d *daemon) start() {
```
vsbogd: I would also made such change to stop() method as well.

Issue #220, opened 2019-02-20
Fully implement "stateless" logic


astroseger:
We should modify ```state_service``` in order to full implement state-less logic described here: https://github.com/singnet/dev-portal/blob/master/docs/concepts/mpe-stateless-client.md

astroseger:
I've removed my initial proposition of simplified state_service because it was obviously wrong, so we should implement which is written in https://github.com/singnet/dev-portal/blob/master/docs/all/mpe/mpe-stateless-client.md

so simple add two new fields:
- oldnonce_signed_amount  last amount which was signed by client with nonce=current_nonce - 1.
- oldnonce_signature - last signature sent by client with nonce = current_nonce - 1. 

it is still very simple...

astroseger:
@raamb 
Can we prioritize this issue? We cannot make SDK without it.

astroseger:
@Dinesh2521 
I would also propose to change authorization strategy, similar to ```control_service``` (include block_number in the message to sign).

astroseger:
There are two issues with the current implementation (#232):
1. It is not backward compatible, and as I believe is possible to do it backward compatible 
2. If ```current_signed_amount``` is 0 then daemon doesn't send ```old_nonce_signed_amount``` even if it is not 0.

It should be noted that we need ```old_nonce_signed_amount``` field **only** when last payment hasn't been claimed in the MPE contract (channel_nonce != block_chain_nonce). Daemon removes payments from the payment storage after it verify that nonce in blockchain has been updated.

It means that we could take ```old_nonce_signed_amount``` directly from payment storage. If there is no payment in payment storage for the given channel_id then we shouldn't send ```old_nonce_signed_amount``` (because we know that client will not need it anyway).

vsbogd:
Taking into account data incompatibility, I would propose to revert last PR and rethink implementation, @Dinesh2521, @raamb  , what do you think?

raamb:
I agree. @Dinesh2521 please revert this PR from main. Once the fixes have been made we can create a new PR

PR #253, opened 2019-04-18
new approach implementation for Fully implement "stateless" logic #220

Dinesh2521:
new approach implementation for Fully implement "stateless" logic #220, as sergy suggested

"It should be noted that we need old_nonce_signed_amount field only when last payment hasn't been claimed in the MPE contract (channel_nonce != block_chain_nonce). Daemon removes payments from the payment storage after it verify that nonce in blockchain has been updated.

It means that we could take old_nonce_signed_amount directly from payment storage. If there is no payment in payment storage for the given channel_id then we shouldn't send old_nonce_signed_amount (because we know that client will not need it anyway)."

we will check the blockchain nonce with storage nonce, if both are different, then we we will retrieve the payment and return it as old_nonce_signed_amount and old_nonce_signature. Since we are retrieving the details from payment directly, we don't need to maintain the old_nonce state explicitly along with channel state storage. 

@vsbogd @astroseger @raamb  can you please review and provide your comments


escrow/escrow.go
```diff
@@ -40,12 +42,30 @@ func (h *lockingPaymentChannelService) PaymentChannelFromBlockChain(key *Payment
...
 	if err != nil {
 		return
 	}
 
+	blockchainChannel, blockchainOk, err := h.blockchainReader.GetChannelStateFromBlockchain(key)
+	if err != nil {
+		return false, errors.New("channel error:" + err.Error())
+	}
+	if !blockchainOk || !storageOk {
+		return false, errors.New("unable to read channel details.")
```
anandrgitnirman: We would know if the error was from block chain or storage no ? , could we add details on where we failed on reading

escrow/state_service.go
```diff
@@ -51,6 +51,29 @@ func (service *PaymentChannelStateService) GetChannelState(context context.Conte
 		return nil, errors.New("only channel signer can get latest channel state")
 	}
 
+	// check if nonce matches with blockchain or not
+	nonceEqual, err := service.channelService.StorageNonceMatchesWithBlockchainNonce(&PaymentChannelKey{ID: channelID})
+	if err != nil {
+		log.WithError(err).Infof("payment data not available in payment storage.")
+	} else if !nonceEqual {
+		// check for payments in the payment storage with current nonce -1, this will happen  cli has issues in claiming process
+		paymentID := fmt.Sprintf("%v/%v", channel.ChannelID, (&big.Int{}).Sub(channel.Nonce, big.NewInt(1)))
```
vsbogd: This is the code duplication of `Payment.ID` method. It is unsafe as if we will change the template of a key there is will be two places to update. 

I would propose adding separate function to get payment id from channel id and channel nonce `PaymentID(string channelID, string channelNonce)` into `payment_channel_api.go` module and implement `Payment.ID` using this function.

escrow/payment_channel_api.go
```diff
@@ -124,6 +124,9 @@ type PaymentChannelService interface {
 
 	//Get Channel from BlockChain
 	PaymentChannelFromBlockChain(key *PaymentChannelKey) (channel *PaymentChannelData, ok bool, err error)
+
+	// verifies whether storage channel nonce is equal to blockchain nonce or not
+	StorageNonceMatchesWithBlockchainNonce(key *PaymentChannelKey) (equal bool, err error)
```
vsbogd: I would not add such specific function into `PaymentChannelService` interface. This function can be implemented in `state_service.go` module instead using 'PaymentChannelService.PaymentChannel()` and `PaymentChannelService.PaymentChannelFromBlockChain()' methods.

escrow/state_service.go
```diff
@@ -51,6 +51,29 @@ func (service *PaymentChannelStateService) GetChannelState(context context.Conte
...
+			// return the channel state with old nonce
+			return &ChannelStateReply{
+				CurrentNonce:         bigIntToBytes(channel.Nonce),
+				CurrentSignedAmount:  bigIntToBytes(channel.AuthorizedAmount),
+				CurrentSignature:     channel.Signature,
+				OldNonceSignedAmount: bigIntToBytes(payment.Amount),
+				OldNonceSignature:    payment.Signature,
+			}, nil
+		}
+		log.WithError(err).Infof("unable to extract old payment from storage" )
```
vsbogd: I would separately detect and handle three cases here:
```go
  if err != nil {
    log.WithError(err).Errorf("unable to extract old payment from storage" )
  }
  if !ok {
    if /* blockchain nonce is not equal to channel nonce, result can be different because of data race */ {
      log.Errorf("old payment is not found in storage, nevertheless local channel nonce is not equal to the blockchain one, channel: %v" channelID)
      return nil, errors.New("channel has different nonce in local storage and blockchain")
    }
  }
  return &ChannelStateReply{
    CurrentNonce:         bigIntToBytes(channel.Nonce),
    CurrentSignedAmount:  bigIntToBytes(channel.AuthorizedAmount),
    CurrentSignature:     channel.Signature,
    OldNonceSignedAmount: bigIntToBytes(payment.Amount),
    OldNonceSignature:    payment.Signature,
  }, nil
```

escrow/payment_channel_storage_test.go
```diff
@@ -81,6 +81,16 @@ func (suite *PaymentChannelStorageSuite) TestGetAll() {
 	assert.Equal(suite.T(), []*PaymentChannelData{channelA, channelB}, channels)
 }
 
+func (suite *PaymentChannelStorageSuite) TestGetChannel() {
```
vsbogd: :+1:

escrow/state_service.go
```diff
@@ -51,6 +51,29 @@ func (service *PaymentChannelStateService) GetChannelState(context context.Conte
 		return nil, errors.New("only channel signer can get latest channel state")
 	}
 
+	// check if nonce matches with blockchain or not
+	nonceEqual, err := service.channelService.StorageNonceMatchesWithBlockchainNonce(&PaymentChannelKey{ID: channelID})
+	if err != nil {
+		log.WithError(err).Infof("payment data not available in payment storage.")
+	} else if !nonceEqual {
```
vsbogd: It is worth to add unit tests for this logic.

escrow/state_service.go
anandrgitnirman: @vsbogd , agreed .
Made this change now

escrow/payment_channel_api.go
anandrgitnirman: fixed this

escrow/state_service.go
anandrgitnirman: fixed this as well

escrow/payment_channel_api.go
anandrgitnirman: @vsbogd , found a minor bug here .. fixing this

vsbogd:
Looks good to me so far, how about unit testing new logic?

escrow/state_service.go
anandrgitnirman: Added

anandrgitnirman:
hi @vsbogd , @raamb 
Now that we have got all the test cases and comments fixed , 
I am also doing a full end to end testing and will confirm if all is well .
( Since this is a critical piece ) , pls do not merge the PR until then

escrow/state_service_test.go
```diff
@@ -82,6 +101,18 @@ func TestGetChannelState(t *testing.T) {
 
 	assert.Nil(t, err)
 	assert.Equal(t, stateServiceTest.defaultReply, reply)
+	//Channel's nonce  =  blockchain nonce + 1
```
vsbogd: I would add it as separate test case to make it possible localize error

escrow/state_service_test.go
```diff
@@ -82,6 +101,18 @@ func TestGetChannelState(t *testing.T) {
...
+	stateServiceTest.defaultChannelData.Nonce = big.NewInt(4)
+	reply, err = stateServiceTest.service.GetChannelState(
+		nil,
+		stateServiceTest.defaultRequest,
+	)
+	assert.Nil(t, err)
+	assert.NotNil(t, reply)
```
vsbogd: I would check that reply contains expected data not just it is not `nil`

escrow/state_service.go
vsbogd: I see that negative cases like "storage error", "channel error" are not covered. Is it big amount of work to cover them?

escrow/escrow_test.go
```diff
@@ -33,6 +33,14 @@ func (p *paymentChannelServiceMock) PaymentChannel(key *PaymentChannelKey) (*Pay
 	return p.data, true, nil
 }
 
+
+func (p *paymentChannelServiceMock) StorageNonceMatchesWithBlockchainNonce(key *PaymentChannelKey)   (equal bool, err error) {
```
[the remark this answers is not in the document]
anandrgitnirman: done, thanks @vsbogd

escrow/state_service_test.go
anandrgitnirman: Makes , Sense done

escrow/state_service_test.go
anandrgitnirman: Done

escrow/state_service.go
anandrgitnirman: I will take a shot at this and let you know if I need anything for help

anandrgitnirman: Apologies missed updating here , this is done and also fixed a minor bug noticed ( we were just logging and not returning the error)

PR #254, opened 2019-04-20
New sign contrto make authorization in state_service the same as in control_service(fix for issue#228)ol service

Dinesh2521:
this fix has 2 of the following changes

change the authorizaion logic in state service similar to the control service. as part of the message to be signed will contain the following fields now instead channel id before
"__get_channel_state" text
channel ID
current block number
we have added fall back to older message format as well, since it may break the interaction with CLI, if the developers are using older CLI versions. this backward compatibility can be removed in future releases.
moved some of the common authorization code to separate package, so that we can extend it to build the authentication service for upcoming changes like operator UI
added basic tests to verify the changes. but we may have to review and add bit more detailed tests while developing it for operator UI

@vsbogd @astroseger @raamb can you please review the changes and let me know if in case of any changes/suggestion

### DIFF
--- a/authutils/auth_service.go
+++ b/authutils/auth_service.go
@@ -18,6 +18,11 @@ import (
 
 // Extracts the signer address from signature given the signature
 // It returns signer address and error. nil error indicates the successful function execution
+
+const (
+	AllowedBlockChainDifference = 5
+)
+
 func GetSignerAddressFromMessage(message, signature []byte) (signer *common.Address, err error) {
 	log := log.WithFields(log.Fields{
 		"message":   blockchain.BytesToBase64(message),
@@ -71,8 +76,8 @@ func CompareWithLatestBlockNumber(blockNumberPassed *big.Int) error {
 		return err
 	}
 	differenceInBlockNumber := blockNumberPassed.Sub(blockNumberPassed, latestBlockNumber)
-	if differenceInBlockNumber.Abs(differenceInBlockNumber).Uint64() > 5 {
-		return fmt.Errorf("difference between the latest block chain number and the block number passed is %v ", differenceInBlockNumber)
+	if differenceInBlockNumber.Abs(differenceInBlockNumber).Uint64() > AllowedBlockChainDifference {
+		return fmt.Errorf("authentication failed as the signature passed has expired")
 	}
 	return nil
 }
--- a/authutils/auth_service.go
+++ b/authutils/auth_service.go
@@ -96,7 +96,7 @@ func CurrentBlock() (*big.Int, error) {
 func VerifyAddress(address common.Address, otherAddress common.Address) error {
 	isSameAddress := otherAddress == address
 	if !isSameAddress {
-		return fmt.Errorf("the payment Address: %s  does not match to what has been registered", blockchain.AddressToHex(&address))
+		return fmt.Errorf("the  Address: %s  does not match to what has been expected / registered", blockchain.AddressToHex(&address))
 	}
 	return nil
 }
--- a/authutils/auth_service.go
+++ b/authutils/auth_service.go
@@ -93,8 +93,8 @@ func CurrentBlock() (*big.Int, error) {
 }
 
 //Check if the payment address/signer passed matches to what is present in the metadata
-func VerifyPaymentAddress(address common.Address, paymentAddress common.Address) error {
-	isSameAddress := paymentAddress == address
+func VerifyAddress(address common.Address, otherAddress common.Address) error {
+	isSameAddress := otherAddress == address
 	if !isSameAddress {
 		return fmt.Errorf("the payment Address: %s  does not match to what has been registered", blockchain.AddressToHex(&address))
 	}
--- a/authutils/auth_service.go
+++ b/authutils/auth_service.go
@@ -65,7 +65,7 @@ func VerifySigner(message []byte, signature []byte, signer common.Address) error
 }
 
 //Check if the block number passed is not more +- 5 from the latest block number on chain
-func CompareBlockNumbers(blockNumberPassed *big.Int) error {
+func CompareWithLatestBlockNumber(blockNumberPassed *big.Int) error {
 	latestBlockNumber, err := CurrentBlock()
 	if err != nil {
 		return err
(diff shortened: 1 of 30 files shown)


vsbogd:
Two general comments:
- Which problem does this PR solve? Do we have any issue or discussion of this approach and its pros and contras? There is a potential issue with including blocknumber into signature: if client and server uses different Ethereum nodes then current block number can be different for them.
- It seems that this PR includes changes from #253 so I think it should be reviewed after #253 is merged

anandrgitnirman:
@vsbogd on
_

> There is a potential issue with including blocknumber into signature: if client and server uses different Ethereum nodes then current block number can be different for them

_
We do a check to see if difference between the block number is not more than +-5


+ @astroseger can you opine on why +-5 ?

authutils/auth_service.go
```diff
@@ -0,0 +1,102 @@
...
+}
+
+//Check if the block number passed is not more +- 5 from the latest block number on chain
+func CompareWithLatestBlockNumber(blockNumberPassed *big.Int) error {
+	latestBlockNumber, err := CurrentBlock()
+	if err != nil {
+		return err
+	}
+	differenceInBlockNumber := latestBlockNumber.Cmp(blockNumberPassed)
+	if -5 <= differenceInBlockNumber || differenceInBlockNumber <= 5 {
```
anandrgitnirman: this is a bug, we should not use cmp to check the difference , will correct this

authutils/auth_service_test.go
```diff
@@ -4,6 +4,7 @@ package authutils
 import (
 	"github.com/magiconair/properties/assert"
```
vsbogd: Why use two different `assert` libraries here: lets remove it and keep `github.com/stretchr/testify/assert`

authutils/auth_service_test.go
```diff
@@ -12,10 +13,10 @@ func TestCompareWithLatestBlockNumber(t *testing.T) {
 	config.Vip().Set(config.EthereumJsonRpcEndpointKey, "https://ropsten.infura.io")
 	currentBlockNum, _ := CurrentBlock()
 	err := CompareWithLatestBlockNumber(currentBlockNum.Add(currentBlockNum, big.NewInt(13)))
-	assert.Equal(t, "difference between the latest block chain number and the block number passed is 13 ", err.Error())
+	assert2.NotNil(t, err)
```
vsbogd: I like previous version which checks message more.

vsbogd: as it checks specific error not generic one

anandrgitnirman: Reverted back to the previous check on exact error message

vsbogd:
@anandsnet , please merge master to this branch to remove changes which were merged with #253

anandrgitnirman:
@vsbogd master was merged on to this branch

vsbogd:
I have skimmed it and I have no objections, but careful review is needed, then it is better to assign someone else. And it probably makes sense to verify API with someone who is going to support it in `snet-cli`.

authutils/auth_service.go
```diff
@@ -0,0 +1,102 @@
...
+}
+
+//Check if the block number passed is not more +- 5 from the latest block number on chain
+func CompareWithLatestBlockNumber(blockNumberPassed *big.Int) error {
+	latestBlockNumber, err := CurrentBlock()
+	if err != nil {
+		return err
+	}
+	differenceInBlockNumber := blockNumberPassed.Sub(blockNumberPassed, latestBlockNumber)
+	if differenceInBlockNumber.Abs(differenceInBlockNumber).Uint64() > 5 {
```
raamb: Lets use a constant instead of 5

authutils/auth_service.go
```diff
@@ -0,0 +1,102 @@
...
+		return fmt.Errorf("difference between the latest block chain number and the block number passed is %v ", differenceInBlockNumber)
```
raamb: Make the error more descriptive without giving out more details something like "Authentication failed as the signature passed in has expired"

authutils/auth_service.go
```diff
@@ -0,0 +1,102 @@
...
+		if err = ethClient.RawClient.CallContext(context.Background(), &currentBlockHex, "eth_blockNumber"); err != nil {
+			log.WithError(err).Error("error determining current block")
+			return nil, fmt.Errorf("error determining current block: %v", err)
+		}
+		return new(big.Int).SetBytes(common.FromHex(currentBlockHex)), nil
+	}
+}
+
+//Check if the payment address/signer passed matches to what is present in the metadata
+func VerifyPaymentAddress(address common.Address, paymentAddress common.Address) error {
```
raamb: Should the AuthUtils be aware of details like PaymentAddress? Isnt it better to rename this method to be a generic one like VerifyAddress. Also the error message should not be specific to a usecase

anandrgitnirman: makes sense , will rename the function and variable name to more generic

escrow/state_service.go
```diff
@@ -80,8 +78,44 @@ func (service *PaymentChannelStateService) GetChannelState(context context.Conte
...
+	// signature verification
+	message := bytes.Join([][]byte{
+		[]byte("__get_channel_state"),
+		channelID.Bytes(),
+		abi.U256(big.NewInt(int64(request.CurrentBlock))),
+	}, nil)
+	signature := request.GetSignature()
+
+	sender, err := authutils.GetSignerAddressFromMessage(message, signature)
+	if err != nil {
```
raamb: Wont the older signature error out?

authutils/auth_service.go
anandrgitnirman: how about  configuration instead ? default it to 5

escrow/state_service.go
anandrgitnirman: if an invalid signature is sent it will error ( which we want to)  , however valid signatures will be interpreted and they dont error  ( re tested this on old snet-cli and confirmed ) but the sender will not match

authutils/auth_service.go
anandrgitnirman: Will do

PR #260, opened 2019-05-06
Issue : 258 - Remove dependency with service metadata file when block chain is disabled

anandrgitnirman:
Remove the dependency with Service MetaData when the Daemon Starts when block chain is disabled
Also make sure the State Service calls dont make block chain calls when the block chain flag is disabled.

### DIFF
--- a/blockchain/serviceMetadata.go
+++ b/blockchain/serviceMetadata.go
@@ -58,13 +58,25 @@ func ServiceMetaData() *ServiceMetadata {
 	} else {
 		//TO DO, have a snetd command to create a default metadata json file, for now just read from a local file
 		// when block chain reading is disabled
-		metadata, err = readServiceMetaDataFromLocalFile("service_metadata.json")
+		if metadata, err = readServiceMetaDataFromLocalFile("service_metadata.json"); err != nil {
+			// If no local file is available be default , then return a dummy metadata to support functionality when blockchain is disabled.
+			log.WithError(err)
+			metadata = &ServiceMetadata{}
+			metadata.getMetaDataWhenBlockChainIsDisabled()
+			return metadata
+		}
+
 	}
 	if err != nil {
 		log.WithError(err).
 			Panic("error on determining service metadata from file")
 	}
 	return metadata
+}
+
+func (s *ServiceMetadata ) getMetaDataWhenBlockChainIsDisabled()  {
+    	s.ServiceType = "grpc"
+    	s.Encoding = "proto"
 }
 
 func readServiceMetaDataFromLocalFile(filename string) (*ServiceMetadata, error) {
--- a/blockchain/serviceMetadata.go
+++ b/blockchain/serviceMetadata.go
@@ -62,7 +62,7 @@ func ServiceMetaData() *ServiceMetadata {
 			// If no local file is available be default , then return a dummy metadata to support functionality when blockchain is disabled.
 			log.WithError(err)
 			metadata = &ServiceMetadata{}
-			metadata.getMetaDataWhenBlockChainIsDisabled()
+			metadata.setDefaultConfiguration()
 			return metadata
 		}
 
@@ -74,7 +74,7 @@ func ServiceMetaData() *ServiceMetadata {
 	return metadata
 }
 
-func (s *ServiceMetadata ) getMetaDataWhenBlockChainIsDisabled()  {
+func (s *ServiceMetadata ) setDefaultConfiguration()  {
     	s.ServiceType = "grpc"
     	s.Encoding = "proto"
 }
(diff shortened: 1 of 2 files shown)


blockchain/serviceMetadata.go
```diff
@@ -58,7 +58,14 @@ func ServiceMetaData() *ServiceMetadata {
 	} else {
 		//TO DO, have a snetd command to create a default metadata json file, for now just read from a local file
 		// when block chain reading is disabled
-		metadata, err = readServiceMetaDataFromLocalFile("service_metadata.json")
+		if metadata, err = readServiceMetaDataFromLocalFile("service_metadata.json"); err != nil {
+			// If no local file is available be default , then return a dummy metadata to support functionality when blockchain is disabled.
+			log.WithError(err)
+			metadata = &ServiceMetadata{}
```
raamb: Lets make the method name more descriptive

anandrgitnirman: Done

blockchain/serviceMetadata.go
```diff
@@ -58,7 +58,14 @@ func ServiceMetaData() *ServiceMetadata {
 	} else {
 		//TO DO, have a snetd command to create a default metadata json file, for now just read from a local file
 		// when block chain reading is disabled
-		metadata, err = readServiceMetaDataFromLocalFile("service_metadata.json")
+		if metadata, err = readServiceMetaDataFromLocalFile("service_metadata.json"); err != nil {
```
vsbogd: I would move condition on loading correct metadata: from file or from blockchain to the `components.ServiceMetaData()`  method. 
In pseudocode:
```go
func (components *Components) ServiceMetaData() *blockchain.ServiceMetadata {
	if components.serviceMetadata != nil {
		return components.serviceMetadata
	}
        if len(config.GetString(config.ServiceMetadataFileKey)) > 0 {
		components.serviceMetadata = readServiceMetaDataFromLocalFile("service_metadata.json")
		return components.serviceMetadata
        }
        if !components.Blockchain().Enabled() {
            components.serviceMetadata = blockchain.ServiceMetaData()
            return components.serviceMetadata
        }
        log.Panic("Blockchain is disabled and no service metadata file is passed in configuration")
}
```

escrow/state_service.go
```diff
@@ -38,6 +40,14 @@ func (service *PaymentChannelStateService) GetChannelState(context context.Conte
 	if err != nil {
 		return nil, errors.New("incorrect signature")
 	}
+	//If block chain is Disabled, then just return the signed Amount as Zero
+	if !config.GetBool(config.BlockchainEnabledKey){
```
vsbogd: Which usecase requires this dummy result?

anandrgitnirman: Even when block chain is set to false in Daemon ( snet-cli or other clients can make calls to state_service) , if we dont check for block chain disabled , then calls were actually being made to get channel information from block chain. Hit upon this scenario when testing 
the below 
1) Block chain was disabled on Daemon 
2) Make calls to Daemon using snet-cli 
( call to state_service should not trigger a call to block chain)

blockchain/serviceMetadata.go
anandrgitnirman: In pseudocode: looks like a typo in the commented line ( this will go on a infinite recursion)
```go
       if !config.GetBool(config.BlockchainEnabledKey) {
           //components.serviceMetadata = blockchain.ServiceMetaData()
             components.serviceMetadata = blockchain.DefaultServiceMetaData() 
            return components.serviceMetadata
        }
      
        log.Panic("Blockchain is disabled and no service metadata file is passed in configuration")
}
```

anandrgitnirman: log.Panic("Blockchain is disabled and no service metadata file is passed in configuration") seems to indicate that if the block chain is disabled, then file is mandatory  ? why enforce this when we just need 2 details ( Service Type  and Encoding ) ?

raamb: Thinking more about this, lets make it mandatory for the service metadata file to be provided by the user. This ensures that if the developer disables the blockchain mode by mistake there is still an explicit step to be performed for the daemon to run. I would request that we clearly display the error message in the log file and on the console about whats happening.

escrow/state_service.go
vsbogd: Ok I see, so `state_service` still tries to call blockchain. 

It is not mandatory, but probably we need to extract public interface from 'blockchain.Processor` and have two implementations: one which deals with Ethereum and another one which doesn't. In such case we would create appropriate instance in `components.go` once and don't add many `if` statements in `Processor` code.

PR #261, opened 2019-05-06
DRAFT PR - Integration Test cases for better coverage  

anandrgitnirman:


snetd/cmd/memory_storage_util.go
```diff
@@ -0,0 +1,109 @@
+package cmd
```
vsbogd: This file looks like absolute copy of `escrow/memory_storage.go`, you can remove it and use `escrow/memory_storage.go` instead. If the problem is how to use private `memoryStorage` struct then answer is: you can use `AtomicStorage` interface instead. `memoryStorage` implements this interface.

scripts/install
```diff
@@ -14,7 +14,9 @@ popd
 pushd vendor/github.com/golang/protobuf
 go install ./protoc-gen-go
 popd
-
 go run resources/installBlockchainDependencies/main.go
 go generate ./...
 popd
+cd escrow
+protoc -I . ./*.proto --go_out=plugins=grpc:.
```
vsbogd: Is it really needed? Line 18 above should generate all necessary protobuf stubs.

escrow/payment_channel_storage.go
```diff
@@ -97,6 +97,14 @@ type BlockchainChannelReader struct {
 	recipientPaymentAddress   func() common.Address
 }
 
+func NewBlockChainChannelReader(a func(channelID *big.Int) (channel *blockchain.MultiPartyEscrowChannel, ok bool, err error), b func() common.Address) *BlockchainChannelReader {
```
vsbogd: If I understand correct this method and `NewChannelPaymentValidatorT` are added to allow creating `BlockchainChannelReader` and `ChannelPaymentValidator` without real Ethereum RPC endpoint. But actually there is blockchain parameter `BlockchainEnabledKey` which disables connection to Ethereum RPC endpoint. And if you set this key to `false` you can use normal `blockchain.Processor` instance to pass it to usual constructors. So these two new constructors are not required actually and I would suggest removing them.

snetd/cmd/serve_test.go
```diff
@@ -1,11 +1,374 @@
...
+				assert.Nil(t,errC)
+				assert.NotNil(t,channel)
+		//	}()
+
+		})
+	}
+}
+
+
+type paymentChannelServiceMock struct {
```
vsbogd: Looks like this struct is not used anywhere.

snetd/cmd/serve_test.go
```diff
@@ -1,11 +1,374 @@
...
+		return nil, p.err
+	}
+
+	return &paymentTransactionMock{
+		channel: p.data,
+		err:     p.err,
+	}, nil
+}
+
+type paymentTransactionMock struct {
```
vsbogd: Looks like this struct is not used anywhere.

escrow/payment_channel_storage.go
vsbogd: Ok, actually I am not correct: 
- disabling `blockchain.Processor` on the configuration level will either lead to disabling payment infrastructure
- disabling `blockchain.Processor` in processor itself will lead to the payment check failures

So it looks like at the moment we cannot test channel closing behavior without `blockchain.Processor`.

vsbogd:
I would suggest make test main loop as simple as:
```go
components := // mocked components
daemon := newDaemon(components)
daemon.start()
// start test service
// call service method which freezes
daemon.stop()
// check that channel is locked
// unfreeze method
// check that channel is notlocked
```
the question is how to form mocked components object?

Appropriate components structure could have following fields values:
```go
type Components struct {
	serviceMetadata            *blockchain.ServiceMetadata // you already formed it in test
	blockchain                 *blockchain.Processor // see below
	etcdClient                 *etcddb.EtcdClient // should be nil, as we going to use mem storage
	etcdServer                 *etcddb.EtcdServer // should be nil
	atomicStorage              escrow.AtomicStorage // NewMemStorage()
	paymentChannelService      escrow.PaymentChannelService // see below
	escrowPaymentHandler       handler.PaymentHandler // real one
	grpcInterceptor            grpc.StreamServerInterceptor // real one
	paymentChannelStateService *escrow.PaymentChannelStateService // real one
	etcdLockerStorage          *escrow.PrefixedAtomicStorage // real one
	providerControlService     *escrow.ProviderControlService // real one
	daemonHeartbeat            *metrics.DaemonHeartbeat // real one
}
```
Only component we should mock is Ethereum. We can do it on two levels:

1. Mock Ethereum client in other words `blockchain` component. It could be done using `SimulatedEthereumEnvironment` in `blockchain/ethereum_test_util.go` which I added to test escrow code / contracts interaction. It is not an obvious task but seems doable. I would implement it by extracting interface from `blockchain.Processor` and implementing it twice: first is already existing `Processor` implementation and second is simulated Ethereum implementation. Pros of such solution is that we will have fully front to back test which covers almost all code of the daemon (except Ethereum client interaction).

2. Mock `paymentChannelService`. It could be done making `paymentChannelServiceMock` from `escrow/escrow_test.go` public. Then its instance can be used as `components` part. 

In both cases to construct proper components object I would add additional components constructor which gets configuration as parameter instead of reading it from file. Then in test one can create crafted configuration to construct correct set of the components for testing.

anandrgitnirman:
Will revisit this and reopen again

PR #270, opened 2019-05-21
will store version as a part of req/resp stats data

prashantramangupta:
will store version as a part of req/resp stats data.


anandrgitnirman:
Generic comment 
Please refer the issues number as well during commit

PR #273, opened 2019-05-29
Additional tests for state returning logic

vsbogd:
Additional changes for #253 .


vsbogd:
it would be more convenient to move tests to using `github.com/stretchr/testify/suite` framework like in `payment_channel_storage_test.go` for example. Then we could do this work automatically before each test in `SetupTest()` method and get rid of some code duplication and `defer`

Issue #283, opened 2019-06-07
Proto definition for Operator UI 


anandrgitnirman:
Refer  #29 for details on design discussion

Issue #29, opened 2018-08-09
Operator UI


ldub:
The SingularityNET daemon should host an admin dashboard for use by service operators.

Should support
* Remote access
* Authentication via password or metamask
* Service status page
* Breakdown of calls and errors by gRPC service method
* Breakdown of calls and errors by client and/or job address
* Blockchain interaction explorer
* Configuration
   * Price, parameters, version, etc
   * Monitoring endpoints and notifications

Stretch goal
* Standalone admin panel that can connect to many instances of snet-daemon (for operators who run many services)

Dinesh2521:
Instead of password based authentication, we can leverage blockchain based signature validation approach, which would be much more secure and easier way.

How it works ?
1. The machine which we are trying to access must have a ethereum account configured for account validation. Only public key will be configured in the machine. Private will be with the service provider or Admin himself
2. Whenever the admin wanted to authenticate himself with the target machine (daemons deployed in here), he must chose a random text and sign it and send the signature to the target machine
3. Target machine validates the signature using the address configured, and allows the admin user to proceed with accessing daemon and perform the intended actions

At present daemon uses receiver account for many blockchain dependent operations. But for Operator UI authentication, we may have to use a different account for the following reasons 
- If a single service provider wanted to deploy multiple services using different account for different services, and manage all the different daemons from single operator Ui interface
- securing the recipient address from exposing its details accidentally by admin users. If we use existing recipient address and if in case the private key is compromised, user may loose all the existing funds in that account as well

Would like implement the approach of having a separate account for operator UI and follow the above mentioned steps for authentication, and keep the daemon recipient address.

raamb:
@arturgontijo @ferrourwheel would like to get your view as well. The intention is to build a standalone app which can manage Daemons. This will make the task of configuring the Daemon very simple as it will be GUI based.

astroseger:
@Dinesh2521 
We already have grpc interface for such type of control in the daemon: ```control_service```. This interface already include blockchain authentication (a little bit different protocol though..)!
In snet-cli we use this interface to actually control the service (see ```snet treasurer```).
We only need to extend this ```control_service``.

This control center is what I initial called the treasurer server... :)

arturgontijo:
This GUI would be great for a service provider to update/control/monitor all his daemons at once.

[...]

PR #284, opened 2019-06-07
Proto for Operator UI 

anandrgitnirman:
Refer #29  on details discussion 
#283

### DIFF
--- a/config/configuration_service.proto
+++ b/config/configuration_service.proto
@@ -93,12 +93,10 @@ message StatusResponse {
 message ConfigurationParameter {
     //Indicates the name of the Configuration
     string name = 1;
-    //Indicates the value of the parameter
-    string value = 2;
     //Used to help fill in details on the UI ( might be used to indicate if a field is mandatory or not)
-    bool mandatory = 3;
+    bool mandatory = 2;
     //Describes the usage of the this configuration
-    string description = 4;
+    string description = 3;
     //Used to indicate the type of the given field ( this will be used as validations while editing a field value)
     enum Type {
         STRING = 0;
@@ -107,20 +105,20 @@ message ConfigurationParameter {
         BOOLEAN =4;
         ADDRESS=5;
     }
-    Type type = 5;
+    Type type = 4;
     //An option to never edit some configurations ( by default, a configuration will be editable )
-    bool editable = 6;
+    bool editable = 5;
     //This is used to bucket possible impact of editing a configuration, for now we just have 2
     enum UpdateAction {
         RESTART_REQUIRED = 0;
         NO_IMPACT = 1;
     }
     //This will indicate if a restart of the Daemon is required or not and is tied to what configuration is changed.
-    UpdateAction restart_daemon = 7;
+    UpdateAction restart_daemon = 6;
     //This is used to group the Daemon configurations ( for example say all block chain related configurations will have the section 'blockchain'
-    string section = 8;
+    string section = 7;
     //This is used to indicate the default value that would have been used unless specified
-    string defaultValue = 9 ;
+    string defaultValue = 8 ;
 }
 
 //Response sent back by Daemon on Read or Successful update
@@ -128,7 +126,15 @@ message ConfigurationResponse {
     //Signature will compromise of the below
     //(block_number,authentication_address,ConfigurationParameter)
     bytes signature = 1;
+    //Holds the current configuration of the Daemon
+    map<string, string> current_configuration = 2;
 
-    //Holds the current configuration of Daemon and is sent back in the response.
-    repeated ConfigurationParameter configuration = 2;
+    //Holds the current static configuration of Daemon ( Various details of of every leaf level attribute )
+    //of Daemon.
+    ConfigurationSchema schema = 3;
+}
+
+//Holds the entire static attributes associated
+message ConfigurationSchema {
+    repeated  ConfigurationParameter details =1;
 }
--- a/config/configuration_service.proto
+++ b/config/configuration_service.proto
@@ -123,8 +123,6 @@ message ConfigurationParameter {
     string defaultValue = 9 ;
 }
 
-
-
 //Response sent back by Daemon on Read or Successful update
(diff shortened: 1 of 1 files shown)


config/configuration_service.proto
```diff
@@ -7,65 +7,128 @@ package config;
...
     // ("_Request_Read", "block_number",authentication_address)
+    //OR
+    //("__Is_Daemon_Processing_Requests", "block_number",authentication_address)
     bytes signature = 1;
 
+    //current block number (signature will be valid only for short time around this block number)
+    uint64 current_block = 2;
+
+    //The address of the person who has rights to make configuration changes on Daemon
+    string user_address = 4;
```
vsbogd: I would propose adding caller authentication message and reusing it in all commands:
```proto
//Caller authentication data
message CallerAuth {
    //Signature will compromise the following data:
    // (method name, current block, user address)
    //Example: ("GetConfigurationScheme", <current_block>) signed by private key
    bytes signature = 1;
    //current block number (signature will be valid only for short time around this block number)
    uint64 current_block = 2;
    //The address of the person who has rights to make configuration changes on Daemon
    string user_address = 3;
}
```

config/configuration_service.proto
```diff
@@ -7,65 +7,128 @@ package config;
...
-}
+    //current block number (signature will be valid only for short time around this block number)
+    uint64 current_block = 3;
+
+    //The address of the person who has rights to make configuration changes on Daemon
+    string user_address = 4;
 
+}
 
 message CommandRequest {
```
vsbogd: Both `ReadRequest` and `CommandRequest` contain only authentication information. If there is no reason to keep them separate we could replace them by single request:
```proto
//Used with methods which don't have other parameters except authentication data
message EmptyRequest {
    //Caller authentication data
    CallerAuth auth = 1;
}
```

config/configuration_service.proto
```diff
@@ -7,65 +7,128 @@ package config;
...
 }
 
 
-
 message Response {
     //Signature will compromise of the below
     //(block_number,authentication_address)
     bytes signature = 1;
+    //Indicates the current processing status ( true means Daemon is processing and false means it is not processing any requests)
+    bool current_processing_status = 2;
```
vsbogd: 1. `Response` sounds general, what do you think about `StatusResponse`
2. I would replace `bool` by enum here, it would make code more self-descriptive:
```proto
message StatusResponse {
    //Signature will compromise of the below
    //(block_number,authentication_address,status)
    bytes signature = 1;

    enum Status {
    	// Daemon has no request in progress
    	IDLE = 0;
    	// Daemon is processing request
    	REQUEST_IN_PROGRESS = 1;
    }

    //Indicates the current processing status
    Status current_processing_status = 2;
}
```

config/configuration_service.proto
```diff
@@ -7,65 +7,128 @@ package config;
...
 }
 
 
+//Represents the configuration of current Daemon, please note every leaf level will be filled with these details.
+message ConfigurationParameter {
```
vsbogd: We could return configuration scheme and configuration separately. In such case configuration is:
```proto
    //Indicates the configuration
    map<string, string> configuration = 2;
```
And scheme is array of `ConfigurationParameter(Scheme)` without value field.

config/configuration_service.proto
anandrgitnirman: I like this , will do

config/configuration_service.proto
anandrgitnirman: Ah, So the original thought was to know if the Demon has stopped processing any request ( speiclaly after the StopProcessingRequests / StartProcessingRequests are called ) 
But I like this better , as we can adopt to more states of Daemon in to future , but have something like HAS_STOPPED_PROCESSING_REQUESTS , IS_READY_TO_PROCESS_REQUESTS ?in the enum

config/configuration_service.proto
anandrgitnirman: Thanks @vsbogd , with this approach , the client will have to keep track of key,value 
and key , Configuration and map them ..
Any additional advantages with the approach you propose ?

config/configuration_service.proto
vsbogd: Yes, one of the enum's advantages is that it can be extended in future.

vsbogd: I don't insist on specific names I proposed, main idea is to have such self-descriptive names.

config/configuration_service.proto
vsbogd: Well, I thought about separating configuration scheme from configuration values mainly because of symmetry. When API has two methods `GetConfiguration` and `UpdateConfiguration` it is natural to be able passing results of `GetConfiguration` to `UpdateConfiguration`. I saw that in our API is not a case because `GetConfiguration` returns values coupled with scheme and I thought that decoupling them will make API more clear and natural. 

Another point: configuration scheme is bigger than configuration values but doesn't be changed frequently. So introducing separate method to get configuration scheme should decrease size of the data which is sent from daemon to client when `GetConfiguration` is called.

On the other hand adding separate method to get scheme has disadvantage as well. With separate method client need a way to find out when configuration scheme is changed. So at the end I think it is worth to add message to represent configuration scheme but still return it in `ConfigurationResponse`:

```protobuf
//Response sent back by Daemon on Read or Successful update
message ConfigurationResponse {
    //Signature will compromise of the below
    //(block_number,authentication_address,configuration)
    bytes signature = 1;

    //Holds the current configuration scheme of Daemon
    ConfigurationScheme scheme = 2;

    //Configuration (key, value) map
    map<string, string> configuration = 3;
}
```

config/configuration_service.proto
vsbogd: Ah ok, I see that you have added `HAS_STOPPED_PROCESSING_REQUESTS` already - good idea :+1:

config/configuration_service.proto
anandrgitnirman: thanks @vsbogd , will do this

Issue #289, opened 2019-06-10
Unit test results are always GREEN despite test failure


vsbogd:
Looks like `grep` call hides results of the tests:

https://github.com/singnet/snet-daemon/blob/d961ebaeeec5059eead6050d106f20c10ab913a2/scripts/test#L13-L22

anandrgitnirman:
@vsbogd , seems the issue is not on the grep but more on the echo used at the line
```sh
then **echo** `GOCACHE=off go test -v ./${d: 0:-1} -coverprofile ${d: 0:-1}.part`;fi 
```
@prashantramangupta will raise a PR, I have tested this as well and able to see failed build after this fix

anandrgitnirman:
reverted back the changes for now ( we will revisit the coverage piece again soon) , we have multiple PR coming up this week and early next week , hence the revert

PR #293, opened 2019-06-11
Fixes Issue: Builds were getting passed even if testcases are getting failed

prashantramangupta:
Fixes Issue #289


anandrgitnirman:
we will revisit this again soon per discussion with team 
refer to #300  on revert fix

Issue #286, opened 2019-06-10
Support Pricing at a method Level on Daemon 


anandrgitnirman:
Daemon should be able to determine the price based on the method invoked 
The method to price details will be stored in the metadata 
Details to follow soon on the json structure

anandrgitnirman:
@vsbogd @astroseger 
the below is what we were thinking of changing the metadata to to support fixed price per method
Let me know if this looks fine to you 
the same will need to be supported on the snet-cli side a well 

```json
"pricing": {
    "price_model": "fixed_price_per_method",
    "details": [
      {
        "service_name": "Calculator",
        "method_pricing": [
          {
            "method_name": "add",
            "price_in_cogs": 2
          },
          {
            "method_name": "sub",
            "price_in_cogs": 1
          },
          {
            "method_name": "div",
            "price_in_cogs": 2
          },
          {
            "method_name": "mul",
            "price_in_cogs": 3
          }
        ]
      },
      {
        "service_name": "Calculator2",
        "method_pricing": [
          {
            "method_name": "add",
            "price_in_cogs": 2
          },
          {
            "method_name": "sub",
            "price_in_cogs": 1
          },
          {
            "method_name": "div",
            "price_in_cogs": 3
          },
          {
            "method_name": "mul",
            "price_in_cogs": 2
          }
        ]
      }
    ]
  },
```

raamb:
@vforvalerio87 @arturgontijo @ferrouswheel

arturgontijo:
Great feature!

ferrouswheel:
Is there any reason we can't map directly from method name to price?

i.e.

```
"pricing": {
    "price_model": "fixed_price_per_method",
    "details": [
      {
        "service_name": "Calculator",
        "method_pricing": {
            "add": 2,
            "sub": 1,
            ...

          },
...
```

raamb:
@ferrouswheel the representation is a bit verbose but one advantage is that we can deserialize the JSON into a standard object irrespective of the method names.

PR #296, opened 2019-06-17
Issue 286 - Changes on Daemon to Support Method Level Pricing 

anandrgitnirman:
#286 
Ensure , Both Fixed Pricing and Method Level Pricing models are Supported in Daemon 
Latest Design Diagram below

pricing


price/fixed_method_pricing.go
```diff
@@ -0,0 +1,42 @@
...
+	"github.com/singnet/snet-daemon/handler"
+	"math/big"
+)
+
+type FixedMethodPrice struct {
+	//Service/Method is the key and value is he price
+	methodToPriceMap map[string]*big.Int
+}
+
+func (priceType FixedMethodPrice) GetPrice(GrpcContext *handler.GrpcStreamContext) (price *big.Int, err error) {
```
vsbogd: Should it be:
```suggestion
func (priceType *FixedMethodPrice) GetPrice(GrpcContext *handler.GrpcStreamContext) (price *big.Int, err error) {
```
?

price/fixed_method_pricing_test.go
```diff
@@ -0,0 +1,37 @@
...
+	"testing"
+
+	"github.com/singnet/snet-daemon/handler"
+	"github.com/stretchr/testify/assert"
+)
+
+
+
+func TestFixedMethodPrice_initPricingData(t *testing.T) {
+	metadata,_ := blockchain.ReadServiceMetaDataFromLocalFile("../service_metadata_method_pricing.json")
```
vsbogd: I would move test data into package and rename it to make clear that it is test data. Even better solution would be using test string instead of external file.

price/pricing.go
```diff
@@ -0,0 +1,71 @@
...
+import (
+	"fmt"
+	"github.com/singnet/snet-daemon/blockchain"
+	"github.com/singnet/snet-daemon/handler"
+	log "github.com/sirupsen/logrus"
+	"math/big"
+	"strings"
+)
+
+type Pricing struct {
```
vsbogd: What do you think about renaming it to `PricingConfig` or `PricingStrategy` (as component name below is `PricingStrategy`)

vsbogd:
It seems logical to rename ` price/iPrice.go` to `price/PriceType.go`

vsbogd:
Overall design looks good to me so far but I doubt we could incorporate all other pricing strategies as easy as well. Anyway for this functionality it looks comparible and fine, for future options we will see.

price/fixed_method_pricing.go
anandrgitnirman: This one was intentionally left without a * ( we would pass a star ideally when a value is to be changed and reflected back at the calling location.
We may need to check the uses of * across daemon for consistency probably start fixing as we touch those classes again.

price/pricing.go
anandrgitnirman: Done

price/fixed_method_pricing.go
```diff
@@ -0,0 +1,42 @@
+package price
```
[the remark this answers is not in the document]
anandrgitnirman: done

escrow/income_test.go
```diff
@@ -16,10 +18,20 @@ func (incomeValidator *incomeValidatorMockType) Validate(income *IncomeData) (er
 	return incomeValidator.err
 }
 
+type  MockPrice  struct{
```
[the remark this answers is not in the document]
anandrgitnirman: Done

escrow/income.go
```diff
@@ -31,17 +32,17 @@ type IncomeValidator interface {
...
-	return &incomeValidator{priceInCogs: priceInCogs}
+func NewIncomeValidator(pricing *price.Pricing) (validator IncomeValidator) {
+	return &incomeValidator{priceStrategy: pricing}
 }
 
 func (validator *incomeValidator) Validate(data *IncomeData) (err error) {
-
-	price := validator.priceInCogs
+//TO DO, the user request information from IncomeData needs to be passed here !!!!
+	price,_ := validator.priceStrategy.GetPrice(data.GrpcContext)
```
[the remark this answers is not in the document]
anandrgitnirman: Done

price/fixed_method_pricing_test.go
anandrgitnirman: I will use test string , ( Dont  want to mix it with the files set up for local testing )

anandrgitnirman: also moved the new json file to a separate testing folder ( useful if someone wants to quickly see how the new pricing json looks like and wants to test this end to end locally )

anandrgitnirman:
Thanks @vsbogd ,
 shall we start checking in the design diagrams created under the respective folders?

vsbogd:
> shall we start checking in the design diagrams created under the respective folders?

I think checking in diagrams is a good idea, not sure should we add it to the same folder or have separate `doc` folder for documentation. What do you think?

anandrgitnirman:
> > shall we start checking in the design diagrams created under the respective folders?
> 
> I think checking in diagrams is a good idea, not sure should we add it to the same folder or have separate `doc` folder for documentation. What do you think?

same folder is my preference ,

price/fixed_method_pricing.go
vsbogd: Ok, I see your point, I am not sure if `Go` compiler will wisely optimize it and make a copy of the  object only when object is changed. I mean in `Go` it is easy to make error like one you have fixed in `cmd.Serve.go`: one developer created method without `*` because object is not modified, and another developer added modifying operation and at the end we have copy of the object and change is not applied to main instance. Unit testing solves it but is our unit test coverage ideal?

Probably we should find some `Golang` best practices on this. Do you have a link?

escrow/income.go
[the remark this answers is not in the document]
vsbogd: unit test is cool :+1:

vsbogd:
Thank you for adding such nice commit comments :+1:

vsbogd:
I have found that commit message style:
```
#<issue>
...<comment>...
```
makes it impossible to navigate to the commit during "blaming", see for example https://github.com/singnet/snet-daemon/blame/master/snetd/cmd/serve.go#L132 It is because github.com automatically replaces `#<issue>` by issue link and it is only link which is available here.

price/fixed_method_pricing.go
vsbogd: Well, I have found couple of links:
- https://flaviocopes.com/golang-methods-receivers/
- https://github.com/golang/go/wiki/CodeReviewComments#receiver-type
- https://golang.org/doc/effective_go.html#pointers_vs_values

Using logic of this links it is preferable to use value receiver here as it is a small object and method doesn't modifies it. I agree it is safer from threading perspective. But I still think that `Golang` ability to silently compile the call of the method using pointer or value can lead to errors like we had in `snet.cmd.serve.go`.

Summary: lets' keep it as is.

PR #297, opened 2019-06-19
Initial code for Operator UI  ( expect more PRs to be raised) 

anandrgitnirman:
Separate  PR will be raised for every method


vsbogd:
I looked at `.proto` seems good to me.

Issue #264, opened 2019-05-07
Viewing Coverage on CircleCI


anandrgitnirman:
Good to see the overall test coverage 
More Documentation here at 
https://circleci.com/docs/2.0/code-coverage/

raamb:
#302 resolves this

PR #302, opened 2019-06-25
test coverage badge

prashantramangupta:
resolves issue #264

### DIFF
--- a/scripts/test
+++ b/scripts/test
@@ -10,5 +10,19 @@ pushd $PARENT_PATH
 ./scripts/lint
 
 # Run integration tests
-GOCACHE=off go test -v ./...
+# Run integration tests
+go get golang.org/x/tools/cmd/cover
+go get github.com/mattn/goveralls
+for d in */ ; do
+  for entry in "$d"/*;do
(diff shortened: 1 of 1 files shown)


vsbogd:
What about hiding failed tests, is this problem resolved?

raamb:
@vsbogd yes that's resolved. Removing the echo fixed that. There was another issue with the script running multiple times per folder which would cause the metrics test to fail (metrics test uses mockable which imposes a fair use policy. So when the tests were run multiple times there would be sporadic failures when mockable throttled). The issue of tests running multiple times has also been fixed now.

@prashantramangupta please add any details I might have missed out

prashantramangupta:
Apart from this, we enabled `Pass secrets to builds from forked pull requests` in circleci Advance Settings.

vsbogd:
Thanks @raamb , @prashantramangupta , cool :+1:. 

BTW @prashantramangupta , is there any way to exclude from coverage autogenerated files like [control_service.pb.go](https://coveralls.io/builds/24196917/source?filename=escrow/control_service.pb.go) as they negatively affect coverage ratio?

prashantramangupta:
Sure, will look into this. There must be some way to avoid autogenerated files.

Issue #268, opened 2019-05-21
Operator UI - Start and Stop Processing Requests in Daemon 


anandrgitnirman:
**Stop taking requests**
At this point , the Daemon will stop taking in any new requests , however any existing requests being processed will be honoured.

**Start taking in requests.**
Restores the normal behavior to start processing new requests again , please note you can do this ONLY after the payment related configurations are updated) or the user had stopped the Daemon from taking in new requests.

PR #304, opened 2019-06-28
Issue : 268 :Operator UI : Enable functionality to start / stop Daemon from taking in new requests 

anandrgitnirman:
#268
An authorised user will be able to perform this through operator UI


handler/interceptors.go
```diff
@@ -137,6 +155,10 @@ func interceptMonitoring(srv interface{}, ss grpc.ServerStream, info *grpc.Strea
 }
 
 func (interceptor *rateLimitInterceptor) intercept(srv interface{}, ss grpc.ServerStream, info *grpc.StreamServerInfo, handler grpc.StreamHandler) error {
+
+	if (interceptor.processRequest == STOP_PROCESING_ANY_REQUEST) {
+		return fmt.Errorf("No requests are currently being processed, please try again later")
```
vsbogd: Probably gRPC error code should be returned, for example:
```suggestion
		return status.New(codes.Unavailable, "No requests are currently being processed, please try again later").Err()
```

vsbogd: Not sure, `Unavailable` suits case well

configuration_service/configuration_service.go
```diff
@@ -51,15 +52,17 @@ func (service ConfigurationService) StopProcessingRequests(ctx context.Context,
 	if err = service.authenticate("_StopProcessingRequests", request.Auth); err != nil {
 		return nil, err
 	}
-	return nil, fmt.Errorf("work in progress")
+	service.broadcast.trigger <- 0
```
vsbogd: I would use constant names here and below instead of `0` and `1`, this will make code more readable

handler/interceptors.go
```diff
@@ -82,6 +83,10 @@ func NewGrpcErrorf(code codes.Code, format string, args ...interface{}) *GrpcErr
 	}
 }
 
+const (
+	START_PROCESSING_ANY_REQUEST = 1
```
vsbogd: It probably makes sense to move these constants into `configuration_service` to use it in `StopProcessingRequests` and `StartProcessingRequests` methods

handler/interceptors.go
```diff
@@ -102,15 +107,28 @@ type PaymentHandler interface {
 
 type rateLimitInterceptor struct {
 	rateLimiter rate.Limiter
+	subscribeToNotification *configuration_service.MessageBroadcaster
```
vsbogd: I suggest renaming to `messageBroadcaster` as `subscribeToNotification` looks unclear for me.

handler/interceptors.go
```diff
@@ -102,15 +107,28 @@ type PaymentHandler interface {
...
+	processRequest int
```
vsbogd: May be `requestProcessingMode` or `requestProcessingState`?

handler/interceptors.go
```diff
@@ -102,15 +107,28 @@ type PaymentHandler interface {
...
+	startStopNotification chan int
```
vsbogd: `requestProcessingNotification` ?

snetd/cmd/components.go
```diff
@@ -33,6 +34,8 @@ type Components struct {
 	daemonHeartbeat            *metrics.DaemonHeartbeat
 	paymentStorage             *escrow.PaymentStorage
 	priceStrategy              *pricing.PricingStrategy
+	configurationService       *configuration_service.ConfigurationService
+	channelBroadcast           *configuration_service.MessageBroadcaster
```
vsbogd: Why don't name it `configurationBroadcaster` or `configurationBroadcastingChannel`

handler/interceptors.go
anandrgitnirman: unavailable seems most appropriate

anandrgitnirman:
@vsbogd , thanks for the feedback , fixed them all

configuration_service/broadcast_message.go
```diff
@@ -32,7 +32,7 @@ func (broadcast *MessageBroadcaster) NewSubscriber() chan int {
 		broadcast.subscribers = make([]chan int,0)
 	}
 	broadcast.subscribers = append(broadcast.subscribers, ch)
-	broadcast.mutex.Unlock()
+	defer broadcast.mutex.Unlock()
```
vsbogd: If error happens between `Lock()` and `defer Unlock()` lines, then defer operator is not be executed and lock is not released. So `defer broadcast.mutex.Unlock()` line should be located just after `broadcast.mutex.Lock()` (same for `Publish` method)

anandrgitnirman: My Bad :(  fixing this now right away

PR #323, opened 2019-07-22
Capture additional user details in Daemon Metrics 

anandrgitnirman:
#322 
Details like thebelow will be set by the respective clients and added to the Daemon metrics 

- type of client calling Deamon ( ex Snet-cli, Dapp, SDK) , these will be set by the respective clients 
- user adderss

### DIFF
--- a/metrics/response_stats.go
+++ b/metrics/response_stats.go
@@ -20,6 +20,7 @@ type CommonStats struct {
 	ClientType          string
 	UserDetails         string
 	UserAgent           string
+	ChannelId           string
 }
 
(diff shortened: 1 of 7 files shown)


metrics/response_stats.go
```diff
@@ -28,6 +32,9 @@ func BuildCommonStats(receivedTime time.Time, methodName string) *CommonStats {
 		ServiceID:           config.GetString(config.ServiceId),
 		ServiceMethod:       methodName,
 		Version:             config.GetVersionTag(),
+		ClientType:          "",
+		UserDetails:         "",
+		UserAgent:           "",
```
raamb: Cant we have channelId in the common stats?

PR #341, opened 2019-07-22
Issue 340: Correct the mockable url of Daemon 

anandrgitnirman:
corrected the Obsolete mockable URL used in Daemon metrics
Removed the usage of default service heartbeat as a mockable url
#340

### DIFF
--- a/metrics/heartbeat_test.go
+++ b/metrics/heartbeat_test.go
@@ -44,23 +44,23 @@ func TestHeartbeatHandler(t *testing.T) {
 	assert.False(t, err != nil)
 	assert.NotNil(t, dHeartbeat, "heartbeat must not be nil")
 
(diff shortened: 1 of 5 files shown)


metrics/heartbeat_test.go
```diff
@@ -116,5 +116,5 @@ func TestSetNoHeartbeatURLState(t *testing.T) {
 func TestValidateHeartbeatConfig(t *testing.T) {
 	err := ValidateHeartbeatConfig()
 	assert.Nil(t, err)
-	assert.Equal(t, false, isNoHeartbeatURL)
+	assert.Equal(t, true, isNoHeartbeatURL)
```
raamb: Is this expected? We are changing the behaviour of the Testcase?

anandrgitnirman: Hi @raam, yes we have changed the default behavior to not have a service heartbeat

Issue #337, opened 2019-07-22
Use the expiry threshold at defined in the Org Level meta data 


raamb:
Expiry threshold old is now part of the Org metadata ,
This change is per the new structure to support single channel re use per Org / Per Group ID
All the services that belong to the same group will share the same expiry threshold

PR #344, opened 2019-07-25
Support the new JSON structure from Organization metadata and Service metadata 

anandrgitnirman:
#337 

 "groups": [
    {
      "group_name": "default_group2",
      "group_id": "99ybRIg2wAx55mqVsA6sB4S7WxPQHNKqa4BPu/bhj+U=",
      "payment": {
        "payment_address": "0x671276c61943A35D5F230d076bDFd91B0c47bF09",
        **"payment_expiration_threshold": 40320,**
        "payment_channel_storage_type": "etcd",
        "payment_channel_storage_client": {
          "connection_timeout": "5s",
          "request_timeout": "3s",
          "endpoints": [
            "http://127.0.0.1:2379"
          ]
        }
      }
    },
}


pricing/PricingStrategy_test.go
```diff
@@ -8,20 +8,25 @@ import (
...
...er_method\",         \"details\": [           {             \"service_name\": \"Calculator\",             \"method_pricing\": [               {                 \"method_name\": \"add\",                 \"price_in_cogs\": 2               },               {                 \"method_name\": \"sub\",                 \"price_in_cogs\": 1               },               {                 \"method_name\": \"div\",                 \"price_in_cogs\": 2               },               {                 \"method_name\": \"mul\",                 \"price_in_cogs\": 3               }             ]           },           {             \"service_name\": \"Calculator2\",             \"method_pricing\": [               {                 \"method_name\": \"add\",                 \"price_in_cogs\": 2               },               {                 \"method_name\": \"sub\",                 \"price_in_cogs\": 1               },               {                 \"method_name\": \"div\",                 \"price_in_cogs\": 3               },               {                 \"method_name\": \"mul\",                 \"price_in_cogs\": 2               }             ]           }         ]       }]     }   ] } "
```
raamb: Can we keep the JSON in a file outside the code and read it in? This way the JSON can be reused across tests as well

service_metadata.json
```diff
@@ -1,26 +1,125 @@
...
+                {
+                  "method_name": "mul",
+                  "price_in_cogs": 2
+                }
+              ]
+            }
+          ]
+        }]
+    },
     {
```
raamb: What IPs are these?

service_metadata.json
```diff
@@ -1,26 +1,125 @@
...
-  "pricing": {
-    "price_model": "fixed_price",
-    "price_in_cogs": 10000000
-  },
+
+  "model_ipfs_hash": "Qmdiq8Hu6dYiwp712GtnbBxagyfYyvUY1HYqkH7iN76UCc",
+  "mpe_address": "0x7E6366Fbe3bdfCE3C906667911FC5237Cc96BD08",
   "groups": [
     {
+      "endpoints": ["http://34.344.33.1:2379","http://34.344.33.1:2389"],
```
raamb: This is a sample metadata file? This is used when the daemon is started with blockchain disabled right? In that case will this not break as it has method level pricing set up assuming ExampleService

blockchain/orginzationMetadata_test.go
```diff
@@ -5,22 +5,25 @@ import (
...
...rgGroupData = "{   \"org_name\": \"organization_name\",   \"org_id\": \"org_id1\",   \"groups\": [     {       \"group_name\": \"default_group2\",       \"group_id\": \"99ybRIg2wAx55mqVsA6sB4S7WxPQHNKqa4BPu/bhj+U=\",       \"payment\": {         \"payment_address\": \"0x671276c61943A35D5F230d076bDFd91B0c47bF09\",         \"payment_expiration_threshold\": 40320,         \"payment_channel_storage_type\": \"etcd\",         \"payment_channel_storage_client\": {           \"connection_timeout\": \"15s\",           \"request_timeout\": \"13s\",           \"endpoints\": [             \"http://127.0.0.1:2379\"           ]         }       }     },      {       \"group_name\": \"default_group\",       \"group_id\": \"99ybRIg2wAx55mqVsA6sB4S7WxPQHNKqa4BPu/bhj+U=\",       \"payment\": {         \"payment_address\": \"0x671276c61943A35D5F230d076bDFd91B0c47bF09\",         \"payment_expiration_threshold\": 40320,         \"payment_channel_storage_type\": \"etcd\",         \"payment_channel_storage_client\": {           \"connection_timeout\": \"15s\",           \"request_timeout\": \"13s\",           \"endpoints\": [             \"http://127.0.0.1:2379\"           ]         }       }     }   ] }"
```
ksridharbabuus: Is it possible to create the json from a shell script using CLI commandlets? so that there will not be any mismatch of the schema across the components.
Until CLI is ready, you can read from a json file.

blockchain/serviceMetadata.go
```diff
@@ -20,44 +20,50 @@ type ServiceMetadata struct {
...
+	MethodName  string `json:"method_name"`
+	PriceInCogs *big.Int    `json:"price_in_cogs"`
+}
+
+
+
 func getRegistryAddressKey() common.Address {
 	address := config.GetRegistryAddress()
 	return common.HexToAddress(address)
 }
```
ksridharbabuus: Make sure that where ever we have address is in checksum format. Some comparisons might fail.

service_metadata.json
anandrgitnirman: removed these

service_metadata.json
anandrgitnirman: we dont need sample metadata any more

blockchain/orginzationMetadata_test.go
anandrgitnirman: Will address this on a seprate PR

etcddb/etcddb_client.go
```diff
@@ -48,20 +51,40 @@ func NewEtcdClientFromVip(vip *viper.Viper,metaData *blockchain.OrganizationMeta
...
+		//Regular http call
+		etcdv3, err = clientv3.New(clientv3.Config{
+			Endpoints:   metaData.GetPaymentStorageEndPoints(),
+			DialTimeout: conf.ConnectionTimeout,
+
+		})
+		if err != nil {
+			return nil,err
+		}
 	}
```
ksridharbabuus: If there is any issue with the certificates the connection will switch to http but the connection string will be https. Yes, etcd will throw an exception.
Is there any better way to check if it is https then use tls otherwise go with the http. We can directly check from the endpoints whether we are going to use http or https based on that we can have the cert validation and then the connection validation.

ksridharbabuus: I can see the checkIfHttps in getTlsConfig, see if we can push the condition outside of the function. where getTlsConfig will be called only if it is https.

Issue #188, opened 2019-01-15
Allow using same ETCD storage for different services


vsbogd:
In order to allow storing data for different services in same ETCD storage we need to add unique prefix for each service's keys. 

Proposed prefix includes:
- organization id
- service id
- network id or registry contract id

Performance cost of having longer keys needs to be investigated.

Simplest way to implement it is to wrap `atomicStorage` to the `PrefixedAtomicStorage` with prefix proposed, see: https://github.com/singnet/snet-daemon/blob/05b4021d74e0516191289d3b95f68d1f94132385/snetd/cmd/components.go#L144-L156 and https://github.com/singnet/snet-daemon/blob/05b4021d74e0516191289d3b95f68d1f94132385/escrow/atomic_storage.go#L31

astroseger:
Actually MPE_address is enough for prefix. channel_id is already implicitly include organization_id and service_id (via group_id).
So I propose to use MPE_address as a prefix. It is enough.

vsbogd:
> So I propose to use MPE_address as a prefix. It is enough.

Storage has more complex keys than `channel_id`. This issue is about has independent namespaces of the keys for different applications, so prefix should reflect application id.

astroseger:
@vsbogd 
Do you mean that there are some keys which do not contains channel_id? 
If it is true then we probably should use group_id as application_id.

vsbogd:
@astroseger, may be there are no such keys right now, but why we should expect that there will not be such cases in future. In order to understand that such solution is correct one need to keep in mind many different low level design details. 

I don't like idea of coupling service specific prefix with channel id. channel_id could be not unique among different services.

anandrgitnirman:
With Single channel re use ( per Organization/per Group) the current key storage shoudl be good, we intend to add the network type ( ropsten/kovan/mainnet) as prefix to the key .With the same etcd cluster being used by all services now, we also wanted to give some authentaticaion  , seprate users will be created for each network and access to the respective keys with the env prefix will be given

PR #354, opened 2019-07-26
add the network name as well on the storage key prefix 

anandrgitnirman:
#188 Add the network name as part of the storage key prefix
With the single channel per Organization per group , the existing key should suffice ( we dont need to

### DIFF
--- a/escrow/payment_channel_storage.go
+++ b/escrow/payment_channel_storage.go
@@ -27,7 +27,7 @@ func NewPaymentChannelStorage(atomicStorage AtomicStorage,metadata *blockchain.S
 			atomicStorage: &PrefixedAtomicStorage{
 				delegate:  atomicStorage,
 				//Add the MPE Network address as the prefix on the key for storage
-				keyPrefix: metadata.MpeAddress+"/payment-channel/storage",
+				keyPrefix: "/"+metadata.MpeAddress+"/payment-channel/storage",
 			},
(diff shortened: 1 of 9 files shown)


anandrgitnirman:
we will need to enforce the configuration below is mandatory 
blockchain_network_selected

escrow/payment_channel_storage.go
```diff
@@ -27,7 +28,8 @@ func NewPaymentChannelStorage(atomicStorage AtomicStorage) *PaymentChannelStorag
 		delegate: &TypedAtomicStorageImpl{
 			atomicStorage: &PrefixedAtomicStorage{
 				delegate:  atomicStorage,
-				keyPrefix: "/payment-channel/storage",
+				//Add the Network as the prefix on the key for storage
```
raamb: What happens if BlockChainNetworkSelected is not set in the config? If I remember correct user can set the registry address directly and not need to set this correct?

anandrgitnirman: the configurtion is now forced and defualt value has been removed , without this config Daemon will not start

raamb: We shouldnt allow the registryaddress and the networkname to be provided as that can lead to incorrect scenarios.

anandrgitnirman: @raamb as discussed , we will now use the MPE address as the storage prefix

PR #355, opened 2019-07-29
Provision for ETCD user name and password as configuration parameters

anandrgitnirman:
These will be used to connect to the ETCD client end points


anandrgitnirman:
we will not user user name and password

PR #370, opened 2019-08-09
INtergrate with metering service to check for allowed free call limit 

anandrgitnirman:


escrow/free_call_payment_handler.go
```diff
@@ -84,3 +96,59 @@ func (h *freeCallPaymentHandler) CompleteAfterError(payment handler.Payment, res
...
 }
 
+func checkIfFreeCallsAreAllowed(username string) (allowed bool, err error) {
+	response,err := sendRequest(nil,config.GetString(config.MeteringEndPoint)+"/usage/freecalls",username)
+	return checkResponse(response)
+}
+
+//Set all the headers before publishing
+func sendRequest(json []byte, serviceURL string,username string) (*http.Response, error) {
+	req, err := http.NewRequest("GET", serviceURL, bytes.NewBuffer(json))
```
ksridharbabuus: Can you please make sure that the services are called on https?

anandrgitnirman: the URL is https

PR #376, opened 2019-08-17
Issue :345 ( Signatature validation for metering calls) 

anandrgitnirman:
Send the Daemon Signature while making metering calls for validation

### DIFF
--- a/authutils/auth_service.go
+++ b/authutils/auth_service.go
@@ -4,14 +4,17 @@ package authutils
 import (
 	"bytes"
 	"context"
+	"crypto/ecdsa"
 	"encoding/hex"
 	"errors"
 	"fmt"
 	"github.com/ethereum/go-ethereum/common"
 	"github.com/ethereum/go-ethereum/crypto"
 	"github.com/singnet/snet-daemon/blockchain"
+	"github.com/singnet/snet-daemon/config"
 	log "github.com/sirupsen/logrus"
 	"math/big"
+	"net/http"
 )
 
 // TODO convert to separate authentication service. VERY MUCH REQUIRED FOR OPERATOR UI AUTHENTICATION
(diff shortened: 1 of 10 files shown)


authutils/auth_service.go
```diff
@@ -105,3 +109,58 @@ func VerifyAddress(address common.Address, otherAddress common.Address) error {
...
+		privateKey, err = crypto.HexToECDSA(privateKeyString)
+		if err != nil {
+			return nil, err
+		}
+	}
+	log.Debug(crypto.PubkeyToAddress(privateKey.PublicKey))
+	return
+}
+
+func signForMeteringValidation(privateKey *ecdsa.PrivateKey, currentBlock *big.Int, prefix string) []byte {
```
ksridharbabuus: Is it possible to include the UserName or UserAccount in the Signature?
So that only the metering happens for the respective user rather than just trusting the UserName parameter to metering service.

ksridharbabuus: Also can you please check that the build is failing.

ksridharbabuus: We may even have to look at the edge scenarios, like what happens if the connectivity to metering service is lost? Do we need to have a retry mechanism or we can ignore like fire and forget?
For now we have a issue and will implement it later.

anandrgitnirman: > Is it possible to include the UserName or UserAccount in the Signature?
> So that only the metering happens for the respective user rather than just trusting the UserName parameter to metering service.

Added now

anandrgitnirman: > Also can you please check that the build is failing.

Fixed this

anandrgitnirman: > > Is it possible to include the UserName or UserAccount in the Signature?
> > So that only the metering happens for the respective user rather than just trusting the UserName parameter to metering service.
> 
> Added now

we do have a re try mechanism , but needs more thought like how many attempts or how long to wait before an other attempt is made  , for now will track this as a seprate issue as advised

anandrgitnirman: #377 https://github.com/singnet/snet-daemon/issues/377

Issue #402, opened 2019-09-17
Retrieve all the channels for a given User, Org and Group 


anandrgitnirman:
According to the new contract here https://github.com/singnet/snet-daemon/blob/master/escrow/state_service.proto

we have added another method `GetAllChannelStates` for fetching all the channels.
the sdk will only be using
```current_nonce
current_signed_amount
amount_deposited
expiry```
from `ChannelStateReply`.

vsbogd:
@anandrgitnirman , is it issue for SDK? May be raise it in SDK repo then?

anandrgitnirman:
hi @vsbogd , So SDK wanted , and we felt Daemon might be able to give a list of all channels if available ( with the latest state) .Advantages 
1) All SDKs refer use this rather than building writing code to get all the channels from block chain 
2) Daemon is one central place to get the state of the channel

Here is what we re planning to do 
1) For the given , sender, group id, receiver ( these attributes are indexed) , get all the events from the block chain ( starting block number till the latest block number) 
2) For every channel created , you will get the channel details ( merge this with the latest ETCD if present ) and then send back the response of the latest state of channels

vsbogd:
Ah ok, somehow I was under impression that it is already implemented in daemon, but I see that only stub is implemented.

anandrgitnirman:
While , the intent seemed fine , this also creates a potential risk if Daemon decides to be a bad actor for example not return all the channels ( skip the ones with a lot of funds still available) .. and force the client to thus add more funds to existing channel even when not required

PR #403, opened 2019-09-18
Draft - Get all channels for the given sender,recepient and group id from block chain 

anandrgitnirman:
Read the open channel Events from block chain ,
Get all the channel Ids created
Get the latest channel state ( merge block chain and etcd state) and send back the latest channel state
#402 

Draft pull request


anandrgitnirman:
refer #402 , this is obsolete

Issue #418, opened 2019-11-07
monitoring_enabled configuration property doesn't disable monitoring


vsbogd:
To remediate https://github.com/singnet/snet-daemon/issues/417 I have tried to set `monitoring_enabled` to false in daemon configuration. But it doesn't work, and it is not used in code anymore.

Its usage was removed by https://github.com/singnet/snet-daemon/commit/47b377ffdca368e2e9896ee9b889c0fb92f0ec3e#diff-86b02e8ab38de7760b5dc8e2acfd5717

Issue #417, opened 2019-11-07
SIGSEGV kills daemon on metrics publishing


vsbogd:
I am trying to call `example-service` in docker and getting `SIGSEGV` on metering message signing. After `panic` daemon dies:
```
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x5ae3e8]

goroutine 142 [running]:
github.com/singnet/snet-daemon/vendor/github.com/ethereum/go-ethereum/crypto.Sign(0xc4202c26e0, 0x20, 0x20, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0)
	/ext-go/1/src/github.com/singnet/snet-daemon/vendor/github.com/ethereum/go-ethereum/crypto/signature_cgo.go:58 +0x68
github.com/singnet/snet-daemon/authutils.GetSignature(0xc420342150, 0x26, 0x26, 0x0, 0x0, 0x0, 0xc420342150)
	/ext-go/1/src/github.com/singnet/snet-daemon/authutils/auth_service.go:118 +0x148
github.com/singnet/snet-daemon/metrics.signForMeteringValidation(0x0, 0xc4205c21a0, 0x132fa92, 0x6, 0xc420805ee0, 0x7, 0xc4203b93f8, 0x7)
	/ext-go/1/src/github.com/singnet/snet-daemon/metrics/utils.go:148 +0x37c
github.com/singnet/snet-daemon/metrics.SignMessageForMetering(0xc42010ab00, 0xc420805ee0)
	/ext-go/1/src/github.com/singnet/snet-daemon/metrics/utils.go:114 +0x13b
github.com/singnet/snet-daemon/metrics.sendRequest(0xc420488000, 0x24a, 0x2d1, 0xc4204ea320, 0x41, 0xc420805ee0, 0x15a6370, 0xc4205a4000, 0x24a)
	/ext-go/1/src/github.com/singnet/snet-daemon/metrics/utils.go:95 +0x29a
github.com/singnet/snet-daemon/metrics.publishJson(0xc420488000, 0x24a, 0x2d1, 0xc4204ea320, 0x41, 0x1, 0xc420805ee0, 0xc4204ea35b)
	/ext-go/1/src/github.com/singnet/snet-daemon/metrics/utils.go:69 +0x64
[...]

PR #420, opened 2019-11-07
Do not force Metering check unless explicitly set in the configs

anandrgitnirman:
#418 
#417

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -72,7 +72,7 @@ const (
 	"ipfs_end_point": "http://localhost:5002/", 
 	"ipfs_timeout" : 30,
 	"max_message_size_in_mb" : 4,
-	"monitoring_enabled": true,
+	"monitoring_enabled": false,
 	"organization_id": "ExampleOrganizationId", 
 	"passthrough_enabled": false,
 	"service_id": "ExampleServiceId", 
--- a/snetd/cmd/components.go
+++ b/snetd/cmd/components.go
@@ -247,7 +247,7 @@ func (components *Components) GrpcInterceptor() grpc.StreamServerInterceptor {
 	}
     //Metering is now mandatory in Daemon
 	metrics.SetDaemonGrpId(components.OrganizationMetaData().GetGroupIdString())
-	if components.Blockchain().Enabled() {
+	if components.Blockchain().Enabled() && config.GetBool(config.MonitoringEnabled) {
         //dont start the  Daemon if it is not correctly configured in the metering system
         meteringUrl := config.GetString(config.MeteringEndPoint)+"/verify"
 		if ok,err := components.verifyMeteringConfigurations(meteringUrl,


vsbogd:
@anandrgitnirman, couple of questions on this:
1. Am I right that FreeCalls doesn't work when monitoring is disabled? I don't see any check on this in PR.
2. Is monitoring used for FreeCalls functionality only? Or metering uses this properties as well? If it is for FreeCalls only may be rename it to FreeCallsEnabled?

anandrgitnirman:
> @anandrgitnirman, couple of questions on this:
> 
> 1. Am I right that FreeCalls doesn't work when monitoring is disabled? I don't see any check on this in PR.
> 2. Is monitoring used for FreeCalls functionality only? Or metering uses this properties as well? If it is for FreeCalls only may be rename it to FreeCallsEnabled?

Sure, this is more readable wil change the name , and yes free calls will not not work monitoring is disabled ( Daemon needs to send stats to metering that a successful free call was made)

raamb:
@anandrgitnirman @vsbogd metering information is needed for beyond just free calls. We will need this for subscriptions and tiered payment models. So I am not sure I agree with the change in this name. For example as a service provider I may choose not to give free trial but only a tiered pricing model in which case we need the metering endpoint.
The key thing to understand with metering is that if not specified most payment models will not be supported.

vsbogd:
@raamb, @anandrgitnirman, things that I got from code and our discussion:
- metering in general is used to get different metrics;
- metering was reused to implement "free calls";
- only place where metering information is signed by private key is free calls handler;
- as far as I see in `master` version client checks that `pvt_key_for_metering` is filled and it is correct on start and fails otherwise;
- in release `2.0.2` private key presence is not checked and daemon starts but gets SIGSEGV on first call received; 

Looks like we have two kinds of metering:
- "pricing metering" - requires message signing;
- general metering - doesn't require message signing.

So I would propose:
- add `enable_metering` option which is enabled by default, but can be disabled; if `enable_metering` is disabled then `metering_end_point` is not required and some pricing strategies doesn't work;
- make metering private key optional;
- rename it to something like `pricing_metering_private_key`;
- add `enable_free_call` option which is disabled by default;
- if service uses payment strategies like "free call" or "tiered pricing" then detect it on start and ask ensure user providing `pricing_metering_private_key` and enable metering;

I still have questions on metering and free calls setup:
1. Which private key should be provided by service publisher? 
2. Should service publisher register public key before using corresponding private key in configuration?
3. Where can I get metering service binaries to setup local environment?

vsbogd:
I addition to previous comment. 

Probably instead of adding `enable_metering` it is better to use existing `monitoring_enabled`.

> rename it to something like pricing_metering_private_key
> add enable_free_call option which is disabled by default;

This two changes will require migration of clients. To eliminate migration we can replace them by:
- keep `pvt_key_for_metering` name
- add `free_calls_enabled` option which is enabled by default;

Then daemon on start could ask service publisher to either provide `pvt_key_for_metering` or disable `free_calls_enabled` option.

raamb:
@vsbogd 

> Looks like we have two kinds of metering:
> "pricing metering" - requires message signing;
> general metering - doesn't require message signing.
This is a nice distinction (which we dont have today). 

Pricing metering
- Needs a private key
- Is needed to support various pricing strategies provided by the foundation
- Can be leveraged to provide stats as well 

General metering
- Is for stats gathering 
- Does not need a private key

> Which private key should be provided by service publisher?
> Should service publisher register public key before using corresponding private key in configuration?

This is done during service curation. When service developers register with us and go thru the KYC process we take them thru this process in the publisher portal. This usecase will be covered as part of the upcoming publisher portal

> Where can I get metering service binaries to setup local environment?

This is not available currently, code is available in the snet-marketplace repo but we need to improve the documentation on that repo. We will provide steps for local setup 


> Which private key should be provided by service publisher?

anandrgitnirman:
So to summarize and the checks we will have 

1) if Free Call is enabled,  then its mandatory to have the following 
   a) Private Key to sign ( Metering service will verify this Signature) 
   b) A Signer address ,  Signature of the free call request wil be verfied with this signer address .
   c) A valid Metering End point 

2) if Only Metering  is enabled 
    All you need is a valid metering end point 

if you do not want Metering , then just disable this flag and the Monitoring Interceptor will be disabled

vsbogd:
> if you do not want Metering , then just disable this flag and the Monitoring Interceptor will be disabled

@anandrgitnirman , do you mean `monitoring_enabled` flag here?

anandrgitnirman:
> monitoring_enabled

Yes , So  flag for Free call and a flag for metering  ( We can re name the monitoring_enabled to metering_enabled)

vsbogd:
Ok, looks like we are on same page here :+1:

anandrgitnirman:
@vsbogd @raamb , there are some additional changes coming in on the service metadata on freecalls, Will request you to defer your review for some more time till I get that in too .. 
so the flags for free call  will be  not be required in Daemon and will be used from service metadata

vsbogd:
> Will request you to defer your review for some more time till I get that in too ..
so the flags for free call will be not be required in Daemon and will be used from service metadata

@anandrgitnirman , should we just close this PR till that time then?

snetd/cmd/components.go
```diff
@@ -267,7 +276,12 @@ func (components *Components) GrpcInterceptor() grpc.StreamServerInterceptor {
 }
 
 //Metering end point authentication is now mandatory for daemon
-func (components *Components) verifyMeteringConfigurations(serviceURL string,groupId string) (ok bool, err error) {
+func (components *Components) verifyAuthenticationSetUpForFreeCall(serviceURL string,groupId string) (ok bool, err error) {
+
+	if _, err = crypto.HexToECDSA(config.GetString(config.PvtKeyForMetering)); err != nil {
+		return false, errors.New("you need a specify a valid private key 'pvt_key_for_metering' given by you as part " +
```
raamb: Change message to "as part of service publication process". We dont need to mention free calls as this is required for any payment model

anandsnet:
All changes are now done and ready for review

README.md
```diff
@@ -236,11 +229,9 @@ The default value set is to 4 (units are in MB ), this is used to configure the
 In case of Large messages , it is recommended to use streaming than setting a very high value on this configuration.
 It is not recommended to set the value more than 4GB
 
-* **monitoring_enabled** (optional; default: `true`) - 
-Enable or Disable monitoring of Requests arrived and response sent back
-
-* **monitoring_svc_end_point** (optional;only applies if `monitoring_enabled` is set to true) - 
-Needs to be a vaild url where the request and response stats are published as part of monitoring
+* **metering_enabled** (optional,default: `false`) -
+This is used to define if metering needs to be enabled or not .You will need to define a valid ` metering_end_point`
```
vsbogd: Am I right that `metering_end_point` description should be added to README.md as well?

README.md
```diff
@@ -257,6 +248,13 @@ see [etcd client configuration](./etcddb#etcd-client-configuration)
 * **payment_channel_storage_server** (optional) - 
 see [etcd server configuration](./etcddb#etcd-server-configuration)
 
+* **pvt_key_for_metering** (optional;only applies if `free_call_enabled` is set to true) 
+This is used for authentication between daemon and the metering service in the context of free calls.
+Daemon will send a signature signed by this private key , metering service will already have the public key corresponding
+to this Daemon ,metering service will ensure that the signer it receives matches the public key configured at its end.
+This is mandatory only when free calls is enabled.
```
vsbogd:
```suggestion
This is mandatory only when free calls are enabled.
```

config/config.go
```diff
@@ -67,12 +67,13 @@ const (
 	"daemon_end_point": "127.0.0.1:8080",
 	"daemon_group_name":"default_group",
 	"daemon_type": "grpc",
+    "free_call_enabled": false,
```
vsbogd: Looks like this is not used anymore.

anandsnet: Yes, we identify if free calls are enabled based on the metadata

README.md
anandrgitnirman: Done now, my bad

Issue #414, opened 2019-11-04
Send the available balance to metering stats 


anandrgitnirman:
As part of stats sent to metering , Daemon will also send the available balance ( net balance = Balance in Channel - Amount consumed) , this is only in the context of paid calls

PR #423, opened 2019-11-18
#414- Publish latest offchain channel state ( Authorized amount) 

anandrgitnirman:
Publishing latest Authorized amount on a given channel ( to show the effective balance on DAPP)

### DIFF
--- a/escrow/payment_handler.go
+++ b/escrow/payment_handler.go
@@ -115,7 +115,7 @@ func PublishChannelStats(payment handler.Payment) (err *handler.GrpcError) {
 		Nonce: paymentTransaction.Channel().Nonce,
 		GroupID:blockchain.BytesToBase64(paymentTransaction.Channel().GroupID[:]),
 	}
-	serviceURL := config.GetString(config.MeteringEndPoint)+"/contract-api/channel/"+channelStats.ChannelId.String()+"/consume"
+	serviceURL := config.GetString(config.MeteringEndPoint)+"/contract-api/channel/"+channelStats.ChannelId.String()+"/balance"
 
 	channelStats.OrganizationID = config.GetString(config.OrganizationId)
 	channelStats.ServiceID = config.GetString(config.ServiceId)
--- a/escrow/payment_handler.go
+++ b/escrow/payment_handler.go
@@ -96,10 +96,13 @@ func (h *paymentChannelPaymentHandler) getPaymentFromContext(context *handler.Gr
 }
 
 func (h *paymentChannelPaymentHandler) Complete(payment handler.Payment) (err *handler.GrpcError) {
-	return paymentErrorToGrpcError(payment.(*paymentTransaction).Commit())
+	if err = paymentErrorToGrpcError(payment.(*paymentTransaction).Commit()); err != nil {
+		PublishChannelStats(payment)
+	}
+	return err
 }
 
-func (h *paymentChannelPaymentHandler) PublishChannelStats(payment handler.Payment) (err *handler.GrpcError) {
+func PublishChannelStats(payment handler.Payment) (err *handler.GrpcError) {
 	if !config.GetBool(config.MeteringEnabled)  {
 		err = handler.NewGrpcErrorf(codes.Internal, "Cannot post latest offline channel state as metering is disabled !!")
 		log.WithError(err.Err())
@@ -120,11 +123,9 @@ func (h *paymentChannelPaymentHandler) PublishChannelStats(payment handler.Payme
 	commonStats := &metrics.CommonStats{
 		GroupID: channelStats.GroupID,UserName:paymentTransaction.Channel().Sender.Hex()}
 	status := metrics.Publish(channelStats,serviceURL,commonStats)
+
 	if !status {
 		log.WithError(fmt.Errorf("Unable to post latest offchain Channel state on contract API End point !! %s",serviceURL))
-
-	}
-	if !status {
 		return handler.NewGrpcErrorf(codes.Internal, "Unable to publish status error")
 	}
 	return nil
--- a/escrow/payment_handler.go
+++ b/escrow/payment_handler.go
@@ -125,7 +125,7 @@ func PublishChannelStats(payment handler.Payment) (err *handler.GrpcError) {
 	status := metrics.Publish(channelStats,serviceURL,commonStats)
 
 	if !status {
-		log.WithError(fmt.Errorf("Unable to post latest offchain Channel state on contract API End point !! %s",serviceURL))
+		log.WithError(status).Errorf("Payment channel payment handler unable to post latest off-chain Channel state on contract API End point %s",serviceURL)
 		return handler.NewGrpcErrorf(codes.Internal, "Unable to publish status error")
 	}
 	return nil
--- a/escrow/payment_handler.go
+++ b/escrow/payment_handler.go
@@ -1,13 +1,16 @@
 package escrow
 
 import (
-	"math/big"
-
+	"fmt"
 	"github.com/ethereum/go-ethereum/common"
 	"google.golang.org/grpc/codes"
 
 	"github.com/singnet/snet-daemon/blockchain"
+	"github.com/singnet/snet-daemon/config"
 	"github.com/singnet/snet-daemon/handler"
+	"github.com/singnet/snet-daemon/metrics"
+	log "github.com/sirupsen/logrus"
+	"math/big"
 )
 
 const (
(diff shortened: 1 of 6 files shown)


handler/interceptors.go
```diff
@@ -130,6 +130,10 @@ type PaymentHandler interface {
 	Complete(payment Payment) (err *GrpcError)
 	// CompleteAfterError completes payment if service returns error.
 	CompleteAfterError(payment Payment, result error) (err *GrpcError)
+
+	//Publishing latest Authorized amount on a given channel ( to show the effective balance on DAPP/platform supported clients)
+	//Force this behaviour for any new handlers on posting the offline channel usage stats
+	PublishChannelStats(payment Payment) (err *GrpcError)
```
vsbogd: Separate method in the interface is not needed. The idea of the interface is to do this work in `Complete` and `CompleteAfterError` methods. I would add `PublishChannelStats` to `paymentChannelPaymentHandler` only and call it from `Complete` and `CompleteAfterError`.

anandrgitnirman: Thanks @vsbogd , changed this , however dont see a need to call this from CompleteAfterError as the ETCD would not have been changed no ?

vsbogd: Hi Anand, thanks, will look at this. 

> however dont see a need to call this from CompleteAfterError as the ETCD would not have been changed no ?

May be, I didn't look at the PR accurately just made the most major comment I had. 

Regarding your previous question how to enforce behavior: my opinion it is not required to enforce it for absolutely all handlers. For example free-call handler doesn't need implementing it. If there is some subclass of handlers to enforce this we can implement sub-interface which inherits `PaymentHandler` (for example ChannelStatsPublishingPaymentHandler) and add this method into it. But I would start from implementing this logic in `Complete` and `CompleteAfterError`.

escrow/payment_handler.go
```diff
@@ -93,7 +96,39 @@ func (h *paymentChannelPaymentHandler) getPaymentFromContext(context *handler.Gr
...
+	if err = paymentErrorToGrpcError(payment.(*paymentTransaction).Commit()); err != nil {
+		PublishChannelStats(payment)
+	}
+	return err
+}
+
+func PublishChannelStats(payment handler.Payment) (err *handler.GrpcError) {
+	if !config.GetBool(config.MeteringEnabled)  {
+		err = handler.NewGrpcErrorf(codes.Internal, "Cannot post latest offline channel state as metering is disabled !!")
+		log.WithError(err.Err())
```
vsbogd: I guess `WithError` will not log anything, though I am not sure, but I would add a message:
```suggestion
		log.WithError(err.Err()).Error("Error in payment handler commit")
```

escrow/payment_handler.go
```diff
@@ -93,7 +96,39 @@ func (h *paymentChannelPaymentHandler) getPaymentFromContext(context *handler.Gr
...
+		AuthorizedAmount:paymentTransaction.payment.Amount,
+		FullAmount:paymentTransaction.Channel().FullAmount,
+		Nonce: paymentTransaction.Channel().Nonce,
+		GroupID:blockchain.BytesToBase64(paymentTransaction.Channel().GroupID[:]),
+	}
+	serviceURL := config.GetString(config.MeteringEndPoint)+"/contract-api/channel/"+channelStats.ChannelId.String()+"/consume"
+
+	channelStats.OrganizationID = config.GetString(config.OrganizationId)
+	channelStats.ServiceID = config.GetString(config.ServiceId)
+	log.Debug(channelStats)
```
vsbogd: I would add some descibing message:
```suggestion
	log.Debugf("Payment handler is publishing channel statistics: %v", channelStats)
```

escrow/payment_handler.go
```diff
@@ -93,7 +96,39 @@ func (h *paymentChannelPaymentHandler) getPaymentFromContext(context *handler.Gr
...
+	commonStats := &metrics.CommonStats{
+		GroupID: channelStats.GroupID,UserName:paymentTransaction.Channel().Sender.Hex()}
+	status := metrics.Publish(channelStats,serviceURL,commonStats)
+
+	if !status {
+		log.WithError(fmt.Errorf("Unable to post latest offchain Channel state on contract API End point !! %s",serviceURL))
```
vsbogd: I believe it should be:
```suggestion
		log.WithError(status).Errorf("Payment channel payment handler unable to post latest off-chain Channel state on contract API End point %s",serviceURL)
```

escrow/payment_handler.go
```diff
@@ -93,7 +96,39 @@ func (h *paymentChannelPaymentHandler) getPaymentFromContext(context *handler.Gr
...
+		FullAmount:paymentTransaction.Channel().FullAmount,
+		Nonce: paymentTransaction.Channel().Nonce,
+		GroupID:blockchain.BytesToBase64(paymentTransaction.Channel().GroupID[:]),
+	}
+	serviceURL := config.GetString(config.MeteringEndPoint)+"/contract-api/channel/"+channelStats.ChannelId.String()+"/consume"
+
+	channelStats.OrganizationID = config.GetString(config.OrganizationId)
+	channelStats.ServiceID = config.GetString(config.ServiceId)
+	log.Debug(channelStats)
+	commonStats := &metrics.CommonStats{
```
vsbogd: Where one can find metrics service protocol description?

escrow/payment_handler.go
```diff
@@ -93,7 +96,39 @@ func (h *paymentChannelPaymentHandler) getPaymentFromContext(context *handler.Gr
 }
 
 func (h *paymentChannelPaymentHandler) Complete(payment handler.Payment) (err *handler.GrpcError) {
-	return paymentErrorToGrpcError(payment.(*paymentTransaction).Commit())
+	if err = paymentErrorToGrpcError(payment.(*paymentTransaction).Commit()); err != nil {
+		PublishChannelStats(payment)
```
vsbogd: Two options here: to fail or to NOT fail client call if metrics were not published. I see the second one is implemented. At the same time in `PublishChannelStats` you return `GrpcError`. Did you implement the second option by intention?

handler/interceptors_test.go
```diff
@@ -96,6 +96,10 @@ func (handler *paymentHandlerMock) CompleteAfterError(payment Payment, result er
 	return handler.completeAfterErrorResult
 }
 
+func (handler *paymentHandlerMock) PublishChannelStats(payment Payment) (err *GrpcError) {
```
vsbogd: I believe this method is not required anymore.

vsbogd:
BTW, @anandrgitnirman do you use [go fmt](https://blog.golang.org/go-fmt-your-code) which is tool to format Golang code according global Golang code style conventions?

metrics/response_stats.go
```diff
@@ -81,9 +91,9 @@ type ResponseStats struct {
 func PublishResponseStats(commonStats *CommonStats, duration time.Duration, err error) bool {
 	response := createResponseStats(commonStats, duration, err)
 	if  strings.Compare(commonStats.PaymentMode,"free-call") ==0  {
-		Publish(response, config.GetString(config.FreeCallEndPoint),commonStats)
+		Publish(response, config.GetString(config.FreeCallEndPoint)+ "/usage/freecalls",commonStats)
```
vsbogd: Where one can find metrics endpoint descriptions?

anandrgitnirman: We are looking at building a Swagger UI to do this for Market place service   , work in progress ..

escrow/payment_handler.go
anandrgitnirman: We are looking at building a Swagger UI to do this for Market place service , work in progress

anandrgitnirman: in the cucrrent state , for now we dont want to fail any client calls when PublishChannelStats fails, however I wanted to keep this option , just in case we need to manage this error in to the future and also have some testable code to see the error when an Error happens

metrics/response_stats.go
vsbogd: @anandrgitnirman , my main concern is that I cannot check that protocol (see comment https://github.com/singnet/snet-daemon/pull/423/files#r348327583) and endpoints are used properly if I don't have protocol description.

anandrgitnirman:
> BTW, @anandrgitnirman do you use [go fmt](https://blog.golang.org/go-fmt-your-code) which is tool to format Golang code according global Golang code style conventions?

Thanks @vsbogd , this is now configured, code hence forth should be formatted per standard

vsbogd:
Looks good to me, except metering protocol details which I cannot check without spec.

Issue #278, opened 2019-06-03
Remove checks to support old signature in state_service 


anandrgitnirman:
To ensure backward compatibility ,  fall back to older signature versions is retained. this is temporary and will need to be cleaned up once all clients  switch over  to the latest proto version of state_service ( snet-cli and dapp) 

```

if channel.Signer != *sender {
		log.Infof("message does not follow the new signature standard. fall back to older signature standard")

		sender, err = authutils.GetSignerAddressFromMessage(bigIntToBytes(channelID), signature)
		if err != nil {
			return nil, errors.New("incorrect signature")
		}
		if channel.Signer != *sender {
			return nil, errors.New("only channel signer can get latest channel state")
		}
```

anandrgitnirman:
#254

PR #424, opened 2019-11-22
#278, remove obsolete code on checking old signatures

anandrgitnirman:
All clients using Daemon ( ex SDK, DAPP and snet-cli) have updated to use the new proto , and hence removing the obsolete code
https://github.com/singnet/snet-daemon/issues/278


raamb:
@anandrgitnirman can you please confirm that this old signature is not being used by any client in particular the DApp in the Metamask flow

anandrgitnirman:
> @anandrgitnirman can you please confirm that this old signature is not being used by any client in particular the DApp in the Metamask flow

Thanks @raamb , just reconfirmed ,Dapp metamask is using the latest signature

Issue #426, opened 2019-11-28
Use freecalls from servcie metadata to check the allowed number of free calls set for a service 


anandrgitnirman:
service metadata now has a place holder to define the number of free calls allowed .Use this data point rather than getting it from metering end point

anandrgitnirman:
this is complete

PR #427, opened 2019-11-28
Issue #426: get number of free calls allowed from the service metadata 

anandrgitnirman:
service metadata now has a place holder to define the number of free calls allowed .Give placeholders to use this field
More test cases for coverage and go fmt formatting

### DIFF
--- a/blockchain/serviceMetadata_test.go
+++ b/blockchain/serviceMetadata_test.go
@@ -1,19 +1,20 @@
 package blockchain
 
 import (
-	"github.com/ethereum/go-ethereum/common"
-	"github.com/singnet/snet-daemon/config"
-	"github.com/stretchr/testify/assert"
 	"math/big"
 	"strings"
 	"testing"
+
+	"github.com/ethereum/go-ethereum/common"
+	"github.com/singnet/snet-daemon/config"
+	"github.com/stretchr/testify/assert"
 )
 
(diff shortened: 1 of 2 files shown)


blockchain/serviceMetadata_test.go
```diff
@@ -1,40 +1,39 @@
...
 
 func TestAllGetterMethods(t *testing.T) {
-	config.Vip().Set(config.FreeCallEndPoint,"http://demo8325345.mockable.io")
+	config.Vip().Set(config.FreeCallEndPoint, "http://demo8325345.mockable.io")
```
vsbogd: BTW, it is not related to this PR, but can this test be executed offline (without internet access)?

vsbogd:
BTW @anandrgitnirman , simple word form like #426 or /issues/426 can be used to designate issue number in commit header. It is much shorter and readable doesn't have `https://github.com/` prefix but still clickable. And it leaves place for short commit description.

blockchain/serviceMetadata_test.go
anandrgitnirman: no , but let me explore on alternatives , we can create an issue to track this

vsbogd: Yes, lets create issue on this. Once I tried to compile daemon on the plane board and found that it doesn't compiles (the problem was in my test of Ethereum functionality). Since that moment I think that offline compilation is a good thing :-)

anandrgitnirman: Sure, done #429

vsbogd:
Looks good to me

Issue #439, opened 2020-01-23
Support free-call tracking in Daemon 


anandrgitnirman:
Idea is to get away from the metering services and have an ability in Daemon  to keep track of free calls made by a given user 

[Link to Design ](https://www.plantuml.com/plantuml/img/bLJ1Rjim3BthAmYVKi3fqExMmv1kRWyRrXsA0Xs68b0jnXOYMuOaj_DX_pxAbcKbQr3iPecFZu_aapVMSUFwHZ7cf5C8sQr1_CALih261olR2fGkkOBEo1Vk4FOuP8pnqcb3SJGCMyeSuBxuDdyMrrqNhsNLBg22OwFR-NJZIX5nQ5vaYOnvJbX_6hdX2YgaT5ytrF_GzSzAbbw6ZLYgJL0kHAeoHhsSEHoJZwCLEY_LzGRjhJOhRQgj5Fcl9x2JCZhPVChVKluP6nDzyR73XO9xZeHYoJz83sbUmJWK8hXMIh-YY1jOBNAft8GYl1yReSiQopt87LXQ1KBDRThU24h9VwA571vXH-nGZWjMk7EG4CTE2Fi6D9Ky3Iss2wlCWxSUdEKzjRnfk5JdfMwIflEiu5sNvH-_wzRLzkPt9ytmcUorl-kRipznec7r1ecwXQ_tF-v0xvPY5e0ThRxXUrpaCd-PzkulLsxey0hcuUOWIClQIsrvGu53qKkrGh-sIWk_5jru_Ncwv9aktKyiIuP8Ld910dK3DiwCqaSvu0dWVOBBYU2qCkh1lzVdmT6CveSZfgo51YxUAXNTv6gP6WakWSwkvcxU0Rtc__CY5Vp7ZQTzcB0-TCA_vMEaq_2ufcdI-4vq5Nucip4cJuIxL0LvgLSkJHpT58mNei5CXwOCCSOss0hwVVu5)

PR #440, opened 2020-01-28
Freecall Support in Daemon - 

anandrgitnirman:
#439 
More test cases are in progress .. but a working version 
Will break this in to smaller PRs for final review

### DIFF
--- a/escrow/free_call.go
+++ b/escrow/free_call.go
@@ -1,0 +1,153 @@
+package escrow
+
+import (
+	"fmt"
+	log "github.com/sirupsen/logrus"
+)
+
+type lockingFreeCallUserService struct {
+	storage *FreeCallUserStorage
+	locker  Locker
+	//todo..see if validator is required here
+	validator      *FreeCallPaymentValidator
+	replicaGroupID func() ([32]byte, error)
+}
+
+func NewFreeCallUserService(
+	storage *FreeCallUserStorage,
+
+	locker Locker,
+	freeCallValidator *FreeCallPaymentValidator, groupIdReader func() ([32]byte, error)) FreeCallUserService {
+
+	return &lockingFreeCallUserService{
+		storage:        storage,
+		locker:         locker,
+		validator:      freeCallValidator,
+		replicaGroupID: groupIdReader,
+	}
+}
+
+func (h *lockingFreeCallUserService) FreeCallUserUsage(key *FreeCallUserKey) (freeCallUser *FreeCallUserData, ok bool, err error) {
+	freeCallUser, ok, err = h.storage.Get(key)
+	if err != nil {
+		return
+	}
+	if !ok {
+		groupId, err := h.replicaGroupID()
+		if err != nil {
+			return
+		}
+		return &FreeCallUserData{UserId: key.userId, OrgId: key.organizationId, ServiceId: key.serviceId, GroupID: groupId, FreeCallsMade: 0}, true, nil
+	}
+	return
+}
+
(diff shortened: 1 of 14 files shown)


escrow/free_call.go
```diff
@@ -0,0 +1,151 @@
...
+		ServiceId: payment.ServiceId, GroupID: blockchain.BytesToBase64(groupId[:])}
+	freeCallUserData, ok, err := h.FreeCallUserUsage(userKey)
+	if err != nil {
+		return nil, NewPaymentError(Internal, "payment freeCallUserData error:"+err.Error())
+	}
+	if !ok {
+		log.Warn("Payment freeCallUserData not found")
+		return nil, NewPaymentError(Unauthenticated, "payment freeCallUserData \"%v\" not found", userKey)
+	}
+	if err != nil {
```
raamb: Do we wait to get a mutex or retry if we dont get the first time?

escrow/free_call.go
```diff
@@ -0,0 +1,151 @@
...
+		return nil, NewPaymentError(Internal, "cannot get mutex for user: %v", userKey)
+	}
+	if !ok {
+		return nil, NewPaymentError(FailedPrecondition, "another transaction on this user: %v is in progress", userKey)
+	}
+	defer func(lock Lock) {
+		if err != nil {
+			e := lock.Unlock()
+			if e != nil {
+				log.WithError(e).WithField("userKey", userKey).WithField("err", err).Error("Transaction is cancelled because of err, but freeCallUserData cannot be unlocked. All other transactions on this freeCallUserData will be blocked until unlock. Please unlock freeCallUserData manually.")
```
raamb: Should we send a notification in this case? if this code executes the user is essentially completely blocked so we should make it easy for the service developer to remedy this

escrow/free_call.go
```diff
@@ -0,0 +1,151 @@
...
+
+func (transaction *freeCallTransaction) Commit() error {
+	defer func(payment *freeCallTransaction) {
+		err := payment.lock.Unlock()
+		if err != nil {
+			log.WithError(err).WithField("transaction", payment).
+				Error("free call user cannot be unlocked because of error." +
+					" All other transactions on this channel will be blocked until unlock." +
+					" Please unlock user for free calls manually.")
+		} else {
```
raamb: Do we need this?

escrow/free_call_api.go
```diff
@@ -0,0 +1,77 @@
...
+import (
+	"fmt"
+	"github.com/ethereum/go-ethereum/common"
+	"math/big"
+
+	"github.com/singnet/snet-daemon/blockchain"
+)
+
+func FreeCallID(userAddress *common.Address, serviceId string) string {
+	return fmt.Sprintf("%v/%v", userAddress, serviceId)
```
raamb: We should include orgID as well right?

escrow/free_call_api.go
```diff
@@ -0,0 +1,77 @@
...
+func (key *FreeCallUserKey) String() string {
+	return fmt.Sprintf("{ID:%v/%v/%v/%v}", key.UserId, key.OrganizationId,
+		key.ServiceId, key.GroupID)
+}
+
+type FreeCallUserData struct {
+	UserId        string
+	ServiceId     string
+	OrgId         string
+	GroupID       [32]byte
```
ksridharbabuus: Keep GroupID as consistent with ServiceId and OrgId.
Keep the OrgId as consistent naming convention where some places it is mentioned as OrganizationId.

escrow/free_call.go
anandrgitnirman: we can do this for the channel locks too

Issue #441, opened 2020-01-28
Standardize the prefix for Storage Keys/path


anandrgitnirman:
With the same ETCD storage likely being shared on mainnet and ropsten , we need to ensure that the storage keys do factor this in well  to make sure the Ropsten entries dont clash in with say entries of mainent or kovan

PR #442, opened 2020-01-28
#441 Storage

anandrgitnirman:
#441 
In the event that the same etcd / storage is used for test enviroment or mainnet . 
Make sure that we handle this in code , so the storage keys have some way to seprate ropsten entries vs mainnet entries 
For example a channel Id =1 can be created on mainnet and also on ropsten , but their status stored in etcd should not clash


vsbogd:
The planned way to keep different storages for different environments in ETCD is to use network id as prefix for the root atomic storage. I mean changing `Components.AtomicStorage()` method to:

```golang
func (components *Components) AtomicStorage() escrow.AtomicStorage {
	if components.atomicStorage != nil {
		return components.atomicStorage
	}

	if config.GetString(config.PaymentChannelStorageTypeKey) == "etcd" {
		storage := components.EtcdClient()
	} else {
		storage := escrow.NewMemStorage()
	}

	// line below is pseudo-code as there is no Processor.getNetworkId() method
	// and we need some custom environment_id for disabled blockchain situation
	environment_id := components.Processor().getNetworkId();
	components.atomicStorage = &PrefixedAtomicStorage {
		delegate = storage,
		keyPrefix: "/" + environment_id,
	}

	return components.atomicStorage
}
```

escrow/payment_storage.go
```diff
@@ -12,13 +13,11 @@ type PaymentStorage struct {
...
-func NewPaymentStorage(atomicStorage AtomicStorage) *PaymentStorage {
+func NewPaymentStorage(atomicStorage AtomicStorage, metadata *blockchain.ServiceMetadata) *PaymentStorage {
 	return &PaymentStorage{
 		delegate: &TypedAtomicStorageImpl{
-			atomicStorage: &PrefixedAtomicStorage{
-				delegate:  atomicStorage,
-				keyPrefix: "/payment/storage",
-			},
+
+			atomicStorage:     NewPrefixedAtomicStorage(atomicStorage, "/payment/storage", metadata.MpeAddress),
```
vsbogd: What is the reason to prefix PaymentStorage by MPE contract address? MPE contract address can be changed even if network is same and we should keep payments for both MPE contracts. For example such thing may happen when MPE contract is updated.

anandrgitnirman: having MPE contract ensures the payments from both these contracts are still stored seprately no ? else the payments can get corrupted with old and new , specially when the key is same no ?

vsbogd: Ok, you are right, I thought that MPE contract address somehow is taken into account in PaymentID but I see it is not true. Let's summarize it (please correct me if I wrong):
- we have three storages: PaymentChannelStorage, PaymentStorage and LockStorage;
- both PaymentChannelStorage and LockStorage are actual for current MPE contract only;
- PaymentStorage in fact can be used with different versions of the MPE contract, for example if MPE contract was upgraded but old payments need to be saved to be claimed;
- we can also keep payments for each MPE separately then to get old payments some daemon instance should be switched temporary for old MPE contract.

So I would suggest add MPE contract address + network_id as environment_id as I explained https://github.com/singnet/snet-daemon/pull/442#issuecomment-579653252 and remove it from all other places.

anandrgitnirman: Thanks @vsbogd , say we have the below scenario 
On MPE contract 1 , 
lets say the LockStorage , the key is on the channel, say channel id 2 , now for some reason , this gets marked locked in the Daemon for MPE contract 1 , ( etcd storage is same ) 

On MPE contract 2 
Lets say a brand new request comes again on a channel say , channe id 2 ( brand new channel) , if we dont distinguish the locking storageof MPE1 and MPE2, then  the old lock from channel 2 on MPE1 will block the request from on Channel 2 from the new MPE contract .no ?

vsbogd: Well, it is not I was going to say. I will try to rephrase the solution suggested taking https://github.com/singnet/snet-daemon/pull/443#discussion_r373450556 into account. We have base AtomicStorage in `components.go` which is used to build other storages: PaymentChannelStorage, PaymentStorage, LockStorage for channels and new one LockStorage for free payments. What I am suggesting:
- get network id on daemon start;
- create new PrefixedStorage using `/<network_id>` as a prefix, use this storage as base for other storages (i.e. return it from `GetAtomicStorage` of `components.go`); this guarantees that storages for different networks never intersect;
- add new component MPESpecificStorage; it is also instance of PrefixedStorage using `/<mpe_contract_address>` as a prefix; as it is also based on storage from previous item the effective prefix is `/<network_id>/<mpe_contract_address>`; this guarantees that storages which are specific for MPE contract version don't intersect;
- use MPESpecificStorage as base for PaymentChannelStorage, PaymentStorage, LockStorage for channels;
- use GetAtomicStorage as base for LockStorage for free payments.

As result the following hierarchy of storages should be built:
```
AtomicStorage(no prefix)
  PrefixedStorage(/<network_id>) returned by `components.GetAtomicStorage()`
    LockStorage for free payments
    PrefixedStorage(/<mpe_contract_address>) - returned by `components.GetMpeSpecificStorage()`
      PaymentChannelStorage
      PaymentStorage
      LockStorage for channels
```

escrow/atomic_storage.go
```diff
@@ -33,6 +33,22 @@ type PrefixedAtomicStorage struct {
 	keyPrefix string
 }
 
+//It is recommended to use this function to create a PrefixedAtomicStorage
+func NewPrefixedAtomicStorage(atomicStorage AtomicStorage, prefix string) *PrefixedAtomicStorage {
+	if err, ok := atomicStorage.(*PrefixedAtomicStorage) ; ok && err == nil {
```
anandrgitnirman: @vsbogd , i dont like this type check very much here , any other thoughts  ? , other parts have shaped well

escrow/atomic_storage.go
```diff
@@ -33,6 +33,23 @@ type PrefixedAtomicStorage struct {
 	keyPrefix string
 }
 
+//It is recommended to use this function to create a PrefixedAtomicStorage
+func NewPrefixedAtomicStorage(atomicStorage AtomicStorage, prefix string) *PrefixedAtomicStorage {
+	 passedStorage, ok := atomicStorage.(*PrefixedAtomicStorage) ;
+	if ok && passedStorage != nil {
```
anandrgitnirman: @vsbogd , dont like this type check here , any other suggestions ?

snetd/cmd/components.go
```diff
@@ -236,7 +255,7 @@ func (components *Components) FreeCallUserService() escrow.FreeCallUserService {
 
 	components.freeCallUserService = escrow.NewFreeCallUserService(
 		components.FreeCallUserStorage(),
-		escrow.NewEtcdLocker(components.AtomicStorage(), components.ServiceMetaData()),
+		escrow.NewEtcdLocker(components.AtomicStorage()),
```
vsbogd: Free-call locks are not MPE specific:
```suggestion
		escrow.NewEtcdLocker(escrow.NewPrefixedAtomicStorage(components.AtomicStorage(), "/free-call/lock")),
```

escrow/lock.go
```diff
@@ -18,18 +17,15 @@ type Locker interface {
...
 	}
 }
 
 // returns new prefixed storage
-func NewLockerStorage(storage AtomicStorage,metadata *blockchain.ServiceMetadata) *PrefixedAtomicStorage {
-	return &PrefixedAtomicStorage{
-		delegate:  storage,
-		keyPrefix:  "/"+metadata.MpeAddress+"/payment-channel/lock",
-	}
+func NewLockerStorage(storage AtomicStorage) *PrefixedAtomicStorage {
```
vsbogd: Taking into account suggested changes in `components.go` this method can be removed.

vsbogd:
BTW, it makes sense to add migration procedure: to copy information from old etcd scheme to the new one. We could add specific key to designate schema version and detect that update is required. What do you think?

escrow/atomic_storage.go
anandrgitnirman: we cant create PrefixedAtomicStorage in components.go unless delegate, is made as Delegate,However we dont need the type casting ;), looked weird to even do on the first place .. have a test case to check the nested paths , i think code looks neater and simpler now , thanks @vsbogd

anandrgitnirman:
> BTW, it makes sense to add migration procedure: to copy information from old etcd scheme to the new one. We could add specific key to designate schema version and detect that update is required. What do you think?

makes sense Vitaly, will create an issue to track this and also cover this as part of our documentation  ,thanks for calling this out

PR #443, opened 2020-01-29
Freecalls Support in Daemon 

anandrgitnirman:
#439 
This PR brings in the support required to store the state inorder to keep track of free call usage

### DIFF
--- a/escrow/free_call_api.go
+++ b/escrow/free_call_api.go
@@ -2,7 +2,27 @@ package escrow
 
 import (
 	"fmt"
+	"math/big"
 )
+
+// To Support Free calls
+type FreeCallPayment struct {
+	//Has the Id of the user making the call
+	UserId string
+	//Service Id .
+	ServiceId string
+	//Organization Id
+	OrganizationId string
+	//Current block number
+	CurrentBlockNumber *big.Int
+	// Signature passed
+	Signature []byte
+}
+
+func (key *FreeCallPayment) String() string {
+	return fmt.Sprintf("{ID:%v/%v/%v}", key.UserId, key.OrganizationId,
+		key.ServiceId)
+}
 
 type FreeCallUserKey struct {
 	UserId         string
@@ -26,7 +46,10 @@ func (data *FreeCallUserData) String() string {
 }
 
 type FreeCallUserService interface {
-	FreeCallUserUsage(key *FreeCallUserKey) (freeCallUser *FreeCallUserData, ok bool, err error)
+	GetFreeCallUserKey(payment *FreeCallPayment) (userKey *FreeCallUserKey, err error)
+
+	//if the user details are not found, send back a new entry to be persisted
+	FreeCallUser(key *FreeCallUserKey) (freeCallUser *FreeCallUserData, ok bool, err error)
 
 	ListFreeCallUsers() (freeCallUsers []*FreeCallUserData, err error)
 
--- a/escrow/free_call_api.go
+++ b/escrow/free_call_api.go
@@ -1,0 +1,50 @@
+package escrow
+
+import (
+	"fmt"
+)
+
+type FreeCallUserKey struct {
+	UserId         string
+	OrganizationId string
+	ServiceId      string
+	GroupID        string
+}
+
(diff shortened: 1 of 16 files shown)


escrow/free_call_storage.go
```diff
@@ -0,0 +1,55 @@
...
+	delegate TypedAtomicStorage
+}
+
+func NewFreeCallUserStorage(atomicStorage AtomicStorage, metadata *blockchain.ServiceMetadata) *FreeCallUserStorage {
+	return &FreeCallUserStorage{
+		delegate: &TypedAtomicStorageImpl{
+			atomicStorage: &PrefixedAtomicStorage{
+				delegate: atomicStorage,
+				//Add the MPE Network address as the prefix on the key for storage
+				keyPrefix: "/" + metadata.MpeAddress + "/free-call-user/storage",
```
vsbogd: Prefixing number of free calls used by MPE contract address means that after MPE contract update all users will have zero number of free calls made. I suggest removing MpeAddress prefix here.

escrow/free_call_api.go
```diff
@@ -0,0 +1,73 @@
...
+func (key *FreeCallPayment) String() string {
+	return fmt.Sprintf("{ID:%v/%v/%v}", key.UserId, key.OrganizationId,
+		key.ServiceId)
+}
+
+type FreeCallUserKey struct {
+	UserId         string
+	OrganizationId string
+	ServiceId      string
+	GroupID        string
```
vsbogd: Are free-calls group specific?

escrow/free_call_storage.go
```diff
@@ -0,0 +1,53 @@
...
+
+type FreeCallUserStorage struct {
+	delegate TypedAtomicStorage
+}
+
+func NewFreeCallUserStorage(atomicStorage AtomicStorage) *FreeCallUserStorage {
+	return &FreeCallUserStorage{
+		delegate: &TypedAtomicStorageImpl{
+			atomicStorage: &PrefixedAtomicStorage{
+				delegate:  atomicStorage,
```
vsbogd: I just found that it means we cannot use MPE contract address as part of environment id in https://github.com/singnet/snet-daemon/pull/442. Network_id should be prefix of the AtomicStorage and we should add MPESpecific storage which has MPE address prefix.

escrow/free_call_api.go
```diff
@@ -0,0 +1,73 @@
...
+// To Support Free calls
+type FreeCallPayment struct {
+	//Has the Id of the user making the call
+	UserId string
+	//Service Id .
+	ServiceId string
+	//Organization Id
+	OrganizationId string
+	//Current block number
+	CurrentBlockNumber *big.Int
```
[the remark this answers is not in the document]
anandsnet: This was there to check if the signature has not expired, historic and all current calls from.dapp sign this way, this will need to remain for backward compatibility with dapp.. new signatures check will be in. A seprate PR

escrow/free_call_api.go
anandsnet: Yes,

escrow/free_call_storage.go
anandsnet: Ok, we will handle this in the other PR ..and standardize this

escrow/free_call_payment_handler_test.go
```diff
@@ -24,22 +25,35 @@ type FreeCallPaymentHandlerTestSuite struct {
...
 	paymentHandler freeCallPaymentHandler
 	privateKey     *ecdsa.PrivateKey
+	key            *FreeCallUserKey
+	data           *FreeCallUserData
+	memoryStorage  *memoryStorage
+	storage        *FreeCallUserStorage
+	metadata       *blockchain.ServiceMetadata
 }
 
+func (suite *FreeCallPaymentHandlerTestSuite) getKey(userId string) *FreeCallUserKey {
```
keshrisohit: @anandrgitnirman groupId is hardcoded it seems?

keshrisohit: it is test

PR #445, opened 2020-02-05
#439 Freecalls - State service checks for free calls and also Authentication checks on Signature

anandrgitnirman:
#439 

Make state service calls to determine how many free calls are left , 
and also Authentication checks on Signature


authutils/auth_service.go
```diff
@@ -83,6 +85,19 @@ func CompareWithLatestBlockNumber(blockNumberPassed *big.Int) error {
...
 }
 
+//Check if the block number ( date on which the token was issued is not more than 1 month)
+func CheckAllowedBlockDifferenceForToken(blockNumberPassed *big.Int) error {
+	latestBlockNumber, err := CurrentBlock()
+	if err != nil {
+		return err
+	}
+	differenceInBlockNumber := blockNumberPassed.Sub(blockNumberPassed, latestBlockNumber)
+	if differenceInBlockNumber.Abs(differenceInBlockNumber).Uint64() > AllowedBlockDifferenceForAuthorizedToken {
```
vsbogd: Why don't use `big.Int.Cmp` instead?

authutils/auth_service.go
```diff
@@ -22,6 +22,8 @@ import (
 
 const (
 	AllowedBlockChainDifference = 5
+	//1 month is approximately 172800 blocks
+	AllowedBlockDifferenceForAuthorizedToken = 172800
```
vsbogd: We may pass it with token instead of hardcoding it.

authutils/auth_service_test.go
```diff
@@ -22,3 +22,16 @@ func TestCompareWithLatestBlockNumber(t *testing.T) {
 	assert.Equal(t, nil, err)
 
 }
+
+func TestCheckAllowedBlockDifferenceForToken(t *testing.T) {
+	config.Vip().Set(config.BlockChainNetworkSelected, "ropsten")
+	config.Validate()
+	currentBlockNum, _ := CurrentBlock()
+	err := CheckAllowedBlockDifferenceForToken(currentBlockNum.Add(currentBlockNum, big.NewInt(2000000)))
```
vsbogd: As far as I understand the logic `CheckAllowedBlockDifferenceForToken` gets a date when token was issued in blocks. Then this date should be less then current date. But in test the date which bigger then current date is passed.
```suggestion
	err := CheckAllowedBlockDifferenceForToken(currentBlockNum.Sub(currentBlockNum, big.NewInt(2000000)))
```

escrow/free_call_api.go
```diff
@@ -17,6 +17,13 @@ type FreeCallPayment struct {
 	CurrentBlockNumber *big.Int
 	// Signature passed
 	Signature []byte
+	//Group ID
+	GroupId string
+	//Auth Token Passed
+	AuthToken []byte
+	//Date on when the token was issued in block number
+	AuthTokenIssueDateBlockNumber *big.Int
```
vsbogd: BTW, why not pass expiration date instead? 
1. It doesn't require some expiration date hardcode on the daemon side. 
2. It makes it clear for client when token is expected to be expired.

authutils/auth_service_test.go
anandrgitnirman: now its Just Cmp

authutils/auth_service.go
anandrgitnirman: got away with this now , we just check if current block < tokens's Expiry block

authutils/auth_service.go
anandrgitnirman: done

authutils/auth_service.go
```diff
@@ -83,6 +83,19 @@ func CompareWithLatestBlockNumber(blockNumberPassed *big.Int) error {
 	return nil
 }
 
+//Check if the block number ( date on which the token was issued is not more than 1 month)
+func CheckIfTokenHasExpired(blockNumberPassed *big.Int) error {
```
vsbogd: I would rename `blockNumberPassed` to `expirationBlock`

escrow/free_call_api.go
```diff
@@ -17,6 +17,13 @@ type FreeCallPayment struct {
 	CurrentBlockNumber *big.Int
 	// Signature passed
 	Signature []byte
+	//Group ID
+	GroupId string
+	//Auth Token Passed
+	AuthToken []byte
+	//Date on when the token was issued in block number
```
vsbogd:
```suggestion
	//Token expiration date in blocks
```

escrow/free_call_payment_handler.go
```diff
@@ -71,17 +71,29 @@ func (h *freeCallPaymentHandler) getPaymentFromContext(context *handler.GrpcStre
 		return
 	}
 
+	authTokenIssueDate, err := handler.GetBigInt(context.MD, handler.FreeCallAuthTokenBlockNumberHeader)
```
vsbogd: Better to rename `authTokenIssueDate` to `authTokenExpiryDate`
And `FreeCallAuthTokenBlockNumberHeader` to `FreeCallAuthTokenExpiryBlockNumberHeader`

escrow/state_service.proto
```diff
@@ -51,3 +51,31 @@ message ChannelStateReply {
...
+    rpc GetFreeCallsAvailable(FreeCallStateRequest) returns (FreeCallStateReply) {}
+}
+
+message FreeCallStateRequest {
+    //Has the user email id
+    string user_id = 1;
+    //signer-token = (user@mail, user-public-key, token_issue_date), this is generated my Market place Dapp
+    //to leverage free calls from SDK/ snet-cli, you will need this signer-token to be downloaded from Dapp
+    bytes token_for_free_call = 2;
+    //Token issue date in Block number
```
vsbogd:
```suggestion
    //Token expiration date in Block number
```

escrow/validation_test.go
```diff
@@ -37,12 +35,22 @@ func SignFreeTestPayment(payment *FreeCallPayment, privateKey *ecdsa.PrivateKey)
...
 		[]byte(config.GetString(config.ServiceId)),
+		[]byte(payment.GroupId),
 		bigIntToBytes(payment.CurrentBlockNumber),
+		payment.AuthToken,
 	}, nil)
 
 	payment.Signature = getSignature(message, privateKey)
 }
 
+func GenerateFreeCallTokenPayment(payment *FreeCallPayment, privateKey *ecdsa.PrivateKey,userAddress common.Address) {
```
vsbogd: Just to confirm am I right that AuthToken includes only:
- user id
- user address
- token expiry block
?
It doesn't include orgId, serviceId and groupId?

resources/blockchain_network_config.json
```diff
@@ -9,7 +9,7 @@
     "network_id":"42"
   },
   "ropsten":{
-    "ethereum_json_rpc_endpoint":"https://ropsten.infura.io",
+    "ethereum_json_rpc_endpoint":"https://ropsten.infura.io/v3/09027f4a13e841d48dbfefc67e7685d5",
```
vsbogd: Why not common SingularityNet project-id?

handler/interceptors.go
```diff
@@ -50,6 +50,10 @@ const (
 	//Will be used to check if the Signature is still valid
 	CurrentBlockNumberHeader = "snet-current-block-number"
 
+	//Place holder to set the free call Auth Token issued
+	FreeCallAuthTokenHeader = "snet-free-call-auth-token-bin"
+    //Block number on when the Token was issued , to track the expiry of the token , which is ~ 1 Month
+    FreeCallAuthTokenBlockNumberHeader =  "snet-free-call-token-issue-block"
```
vsbogd:
```suggestion
    FreeCallAuthTokenBlockNumberHeader =  "snet-free-call-token-expiry-block"
```

escrow/validation_test.go
anandrgitnirman: this is correct, but when you make a call to state_service / actuall call to Daemon , you need to sign using orgid, serviceid and groupid , ( the other signture) . .. do you recommend adding this on token too?

resources/blockchain_network_config.json
anandrgitnirman: kept this for testing alone ..

authutils/auth_service.go
anandrgitnirman: Done

vsbogd: Excuse me but I don't see the change? May be you didn't push it?

escrow/validation_test.go
vsbogd: > do you recommend adding this on token too?

Not at all. It depends on requirements should user get new token for each service or can use one for all of them. My opinion the last will be simpler for user and for implementation so lets leave it as is.

authutils/auth_service.go
anandrgitnirman: Apologies, I missed changing the name of the parameter

PR #450, opened 2020-02-17
Freecalls - Remove configs for free_call

anandrgitnirman:
#439 
Previously free-calls were used for metering , now Damon maintains a record/storage of all free call, this is to clean up the obsolete code in Daemon


anandrgitnirman:
> Is it possible to say that the methods are deprecated than removing it directly from the Daemon.
> Backward compatibility is important.

tested with Dapp, we are good ,

Issue #452, opened 2020-02-22
Increase code coverage of free call flow 


anandrgitnirman:
additional test cases  for free call state service

PR #453, opened 2020-02-22
#452 Freecalls Test case covergae 

anandrgitnirman:
Increase coverage and fix test cases 
#452

### DIFF
--- a/escrow/free_call_state_service_test.go
+++ b/escrow/free_call_state_service_test.go
@@ -1,0 +1,115 @@
+package escrow
+
+import (
+	"bytes"
+	"crypto/ecdsa"
+	"github.com/ethereum/go-ethereum/common"
+	"github.com/ethereum/go-ethereum/crypto"
+	"github.com/singnet/snet-daemon/blockchain"
+	"github.com/singnet/snet-daemon/config"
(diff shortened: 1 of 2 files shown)


escrow/free_call_state_service_test.go
```diff
@@ -0,0 +1,115 @@
...
+	suite.storage = NewFreeCallUserStorage(suite.memoryStorage)
+	suite.service = NewFreeCallUserService(suite.storage,
+		NewEtcdLocker(suite.memoryStorage, suite.serviceMetaData), func() ([32]byte, error) { return suite.orgMetaData.GetGroupId(), nil },
+		suite.serviceMetaData)
+	suite.stateService = NewFreeCallStateService(suite.orgMetaData, suite.serviceMetaData, suite.service, suite.freeCallPaymentValidator)
+
+
+}
+
+func (suite *FreeCallStateServiceSuite) TestGetFreeCallsAvailable() {
```
raamb: Does this also test the decrease in the number of free calls post service invocation?

anandrgitnirman: this test case is for state service , for increase in the free call count, we have a seprate test case

Issue #454, opened 2020-02-22
Feature to also allow only certain users on making calls to Daemon


anandrgitnirman:
Feature in Daemon to restrict access to only certain users , this feature is useful,when you are in a test environment and dont want everyone to make requests to your service.

PR #455, opened 2020-02-22
#454 Restric users who can make calls during curation process 

anandrgitnirman:
As part of Publisher portal review / curation process , we need to protect the service provider 's serivce in test environment from being called by anyone, only Authroized users can make calls , when the curation process in progress.


escrow/validation.go
```diff
@@ -206,3 +214,15 @@ func bigIntToBytes(value *big.Int) []byte {
 func bytesToBigInt(bytes []byte) *big.Int {
 	return (&big.Int{}).SetBytes(bytes)
 }
+
```
raamb: Lets make it a generic error message instead of specifying that the service is under curation

config/config.go
```diff
@@ -196,10 +201,23 @@ func Validate() error {
...
+	}
 	return validateMeteringChecks()
 }
+func curationChecks() (error) {
+	if GetBool(IsCurationInProgress) {
+		if !common.IsHexAddress(GetString(CurationAddressForValidation)) {
+			return errors.New("A Valid Address needs to be specified to ensure that, only this user can make calls in test during curation process")
+		}
+	}
+	if GetBool(IsCurationInProgress) && GetString(BlockChainNetworkSelected)=="mainnet" {
```
raamb: Typo in message. Lets rephrase it as "Service cannot be curated while set up against Ethereum mainnet"

Issue #456, opened 2020-02-24
Commands to Manage operations on free call users


anandrgitnirman:
With free call usage now being maintained by Daemon , you should ideally give the Service providers option to view the list of all free call users for your service,
 also give an option to release any locked state of the user
and an option to re set the counter ( i.e allow the user to make more calls)

vsbogd:
> We should also give an option to release any locked state of the user
@anandrgitnirman could you please explain it further. What kind of locked state do you mean?

anandrgitnirman:
> kind of locked state do you mea

while updating the state of the free call, we use the same approach as that of channels, first acquire a lock and then once the service call is successful , update the calls made and release the lock . We have noticed from past experience that a channel can get locked ( since the concept is similar on how we use locks for free call usage update, there can be potential risk that the user gets locked too  LockStorage) ,

vsbogd:
BTW, for the free-calls feature it would be simpler to decrement free-calls counter atomically before returning the result. In such case locking is not required. What do you think?

anandrgitnirman:
> BTW, for the free-calls feature it would be simpler to decrement free-calls counter atomically before returning the result. In such case locking is not required. What do you think?

when service provider changes the free calls allowed , it becomes tricky to manager this , when we decrement no ? .. In which case we could say .. increment the number of free calls made before returning the result ., however . we also wanted to mirror the behaviour of paid calls ( for example when a request is currently in progress, we lock the  channel state offchain and other requests that did not get the lock can't proceed further ) .. similarl will be the case for free calls

vsbogd:
> however . we also wanted to mirror the behaviour of paid calls ( for example when a request is currently in progress, we lock the channel state offchain and other requests that did not get the lock can't proceed further ) .. similarl will be the case for free calls

Well, I don't see a clear reason to reproduce payment channel calls behavior. I didn't find a way to apply payments without locking. It is the only reason why lock is used in `escrow` calls. In `free-calls` case as you mentioned we can increment number of calls made. We already have `CompareAndSwap` operation in `AtomicStorage` interface to implement such things easily.

I don't insist, my intention was to clarify whether fixing this issue is much simpler than implementing non-blocking `free-calls`. If it is not true than implementing non-blocking algorithm for `free-calls` makes more senses.

PR #457, opened 2020-02-24
#456 : Commands to Manage operations on free call users

anandrgitnirman:
View the full list of users who have made free calls to your service and also the usage ( i.e how many free calls)


snetd/cmd/free_call_users.go
```diff
@@ -0,0 +1,176 @@
...
+// initializes and returns the new unlock user of free calls command object
+func newFreeCallUserUnLockCommandCommand(cmd *cobra.Command, args []string, pComponents *Components) (command Command, err error) {
+	userID, err := getUserId(cmd)
+	if err != nil {
+		return
+	}
+	command = &freeCallUserUnLockCommand{
+		lockStorage: pComponents.LockerStorage(),
+		userId:      userID,
+		component:   pComponents,
```
vsbogd: Why don't keep organization metadata instead of `pComponents`?

snetd/cmd/free_call_users.go
```diff
@@ -0,0 +1,176 @@
...
+func newFreeCallResetCountCommand(cmd *cobra.Command, args []string, pComponents *Components) (command Command, err error) {
+	userID, err := getUserId(cmd)
+	if err != nil {
+		return
+	}
+	command = &freeCallUserResetCountCommand{
+		lockStorage: pComponents.LockerStorage(),
+		userStorage: pComponents.FreeCallUserStorage(),
+		userId:      userID,
+		component:   pComponents,
```
vsbogd: Why don't keep organization metadata instead of pComponents?

snetd/cmd/free_call_users.go
```diff
@@ -0,0 +1,176 @@
...
+}
+
+// initializes and returns the new unlock user of free calls command object
+func newFreeCallUserUnLockCommandCommand(cmd *cobra.Command, args []string, pComponents *Components) (command Command, err error) {
+	userID, err := getUserId(cmd)
+	if err != nil {
+		return
+	}
+	command = &freeCallUserUnLockCommand{
+		lockStorage: pComponents.LockerStorage(),
```
vsbogd: As far as I understand `FreeCallUserService` uses another lock storage: https://github.com/singnet/snet-daemon/blob/06bb92bb0ffa4811cbac50e1809f5f391657f62f/snetd/cmd/components.go#L258

snetd/cmd/free_call_users.go
```diff
@@ -0,0 +1,176 @@
...
+func (command *freeCallUserUnLockCommand) unlockFreeCallUser() (err error) {
+	key := &escrow.FreeCallUserKey{}
+	key.UserId = freeCallUserId
+	key.OrganizationId = config.GetString(config.OrganizationId)
+	key.ServiceId = config.GetString(config.ServiceId)
+	key.GroupID = command.component.OrganizationMetaData().GetGroupIdString()
+	// check whether the key exists or not
+	_, ok, err := command.lockStorage.Get(key.String())
+	if !ok {
+		fmt.Printf("Error: Free Call user %s not found\n", key.String())
```
vsbogd:
```suggestion
		fmt.Printf("Error: Free Call lock for user %s is not found\n", key.String())
```

snetd/cmd/free_call_users.go
```diff
@@ -0,0 +1,176 @@
...
+	err = command.lockStorage.Delete(key.String())
+	if err != nil {
+		fmt.Printf("Error: Unable to unlock the user -%s\n", key.String())
+		return
+	}
+	fmt.Printf("Success: User %s unlocked\n", key.String())
+	return
+}
+
+// release the lock on the user with the given user id
```
vsbogd:
```suggestion
// reset free locks counter for a given user id
```

snetd/cmd/free_call_users.go
```diff
@@ -0,0 +1,176 @@
...
+}
+
+// initializes and returns the new reset count for free call user command object
+func newFreeCallResetCountCommand(cmd *cobra.Command, args []string, pComponents *Components) (command Command, err error) {
+	userID, err := getUserId(cmd)
+	if err != nil {
+		return
+	}
+	command = &freeCallUserResetCountCommand{
+		lockStorage: pComponents.LockerStorage(),
```
vsbogd: Is not used as far as I see:
```suggestion
```

snetd/cmd/free_call_users.go
vsbogd: I would suggest adding new component `FreeCallLockStorage` and use it here and in `components.FreeCallUserService` method

snetd/cmd/free_call_users.go
anandrgitnirman: makes sense , we can get away from Components here

PR #458, opened 2020-02-27
#454 - Feature to allow only certain users to make calls to Daemon 

anandrgitnirman:
#454

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -60,7 +60,7 @@ const (
 	//This defaultConfigJson will eventually be replaced by DefaultDaemonConfigurationSchema
 	defaultConfigJson string = `
 {
-    "allowed_user_flag" :false,
+	"allowed_user_flag" :false,
 	"auto_ssl_domain": "",
 	"auto_ssl_cache_dir": ".certs",
 	"blockchain_enabled": true,
--- a/config/config.go
+++ b/config/config.go
@@ -213,7 +213,7 @@ func allowedUserConfigurationChecks() error {
 		if GetString(BlockChainNetworkSelected) == "main" {
 			return fmt.Errorf("service cannot be restricted to certain users when set up against Ethereum mainnet,the flag %v is set to true", AllowedUserFlag)
 		}
-		if err := setAllowedUsers(); err != nil {
+		if err := SetAllowedUsers(); err != nil {
 			return err
 		}
 	}
@@ -340,7 +340,7 @@ func IsAllowedUser(address *common.Address) bool {
 }
 
 //Set the list of allowed users
-func setAllowedUsers() (err error) {
+func SetAllowedUsers() (err error) {
 	users := vip.GetStringSlice(AllowedUserAddresses)
 	if users == nil || len(users) == 0 {
 		return fmt.Errorf("a valid Address needs to be specified for the config %v to ensure that, only these users can make calls", AllowedUserAddresses)
--- a/config/config.go
+++ b/config/config.go
@@ -3,6 +3,7 @@ package config
 import (
 	"errors"
 	"fmt"
+	"github.com/ethereum/go-ethereum/common"
 	"math/big"
 	"net"
 	"net/url"
@@ -25,13 +26,16 @@ const (
 	BlockChainNetworkSelected = "blockchain_network_selected"
 	BurstSize                 = "burst_size"
 	ConfigPathKey             = "config_path"
-
+    CurationAddressForValidation   = "curation_address_for_validation"
 	DaemonGroupName                = "daemon_group_name"
 	DaemonTypeKey                  = "daemon_type"
 	DaemonEndPoint                 = "daemon_end_point"
 	ExecutablePathKey              = "executable_path"
 	IpfsEndPoint                   = "ipfs_end_point"
 	IpfsTimeout                    = "ipfs_timeout"
+	//If this flag is set to true , then request from Daemon will only be taken from a
+	//predefined address
+	IsCurationInProgress           = "is_curation_in_progress"
 	LogKey                         = "log"
 	MaxMessageSizeInMB             = "max_message_size_in_mb"
 	MeteringEnabled                = "metering_enabled"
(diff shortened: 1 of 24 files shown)


config/config.go
```diff
@@ -18,24 +18,21 @@ import (
...
-	AuthenticationAddress     = "authentication_address"
-	AutoSSLDomainKey          = "auto_ssl_domain"
-	AutoSSLCacheDirKey        = "auto_ssl_cache_dir"
-	BlockchainEnabledKey      = "blockchain_enabled"
-	BlockChainNetworkSelected = "blockchain_network_selected"
-	BurstSize                 = "burst_size"
-	ConfigPathKey             = "config_path"
-    CurationAddressForValidation   = "curation_address_for_validation"
+	AllowedUserFlag                = "allowed_user_flag"
+	AllowedUsers                   = "allowed_users"
```
ksridharbabuus: Shall we call it as AllowedUserAddresses in consistent with other variables.

config/config.go
```diff
@@ -18,24 +18,21 @@ import (
...
-	//Contains the Authentication address that will be used to validate all requests to update Daemon configuration remotely through a user interface
-	AuthenticationAddress     = "authentication_address"
-	AutoSSLDomainKey          = "auto_ssl_domain"
-	AutoSSLCacheDirKey        = "auto_ssl_cache_dir"
-	BlockchainEnabledKey      = "blockchain_enabled"
-	BlockChainNetworkSelected = "blockchain_network_selected"
-	BurstSize                 = "burst_size"
-	ConfigPathKey             = "config_path"
-    CurationAddressForValidation   = "curation_address_for_validation"
+	AllowedUserFlag                = "allowed_user_flag"
```
ksridharbabuus: Do we need to have the additional flag, cant we manage with the allowedusernames field. Tough for admin to update two configurations...

config/config.go
```diff
@@ -201,19 +198,25 @@ func Validate() error {
...
-	if GetBool(IsCurationInProgress) && !common.IsHexAddress(GetString(CurationAddressForValidation)) {
-		return fmt.Errorf("a valid Address needs to be specified for the config %v to ensure that, only this user can make calls",CurationAddressForValidation)
-	}
 
+//Feature in Daemon to restrict access to only certain users , this feature is useful,when you are
+//in a test environment and dont want everyone to make requests to your service.
+//Since this was flag was introduced to restrict users while in testing mode, we dont want this configuration
+//to be mistakenly set on mainnet
+func allowedUserConfigurationChecks() error {
+	if GetBool(AllowedUserFlag) {
```
ksridharbabuus: With the above comment, we can get the condition with the allowedUserNames itself...

anandrgitnirman: we wanted to use a flag to allow  a feature to be used, rather than derive the behaviour based on a field being populated , the code looks slightly better with the flag checks, however agree we can derive if the feature is On/off based on the value of the allowedUsers configuration

config/config.go
```diff
@@ -63,6 +60,7 @@ const (
 	//This defaultConfigJson will eventually be replaced by DefaultDaemonConfigurationSchema
 	defaultConfigJson string = `
 {
+    "allowed_user_flag" :false,
```
vsbogd:
```suggestion
	"allowed_user_flag" :false,
```

handler/interceptors.go
```diff
@@ -372,9 +375,47 @@ func GetSingleValue(md metadata.MD, key string) (value string, err *GrpcError) {
 // NoOpInterceptor is a gRPC interceptor which doesn't do payment checking.
 func NoOpInterceptor(srv interface{}, ss grpc.ServerStream, info *grpc.StreamServerInfo,
 	handler grpc.StreamHandler) error {
+	if config.GetBool(config.AllowedUserFlag){
```
vsbogd: Why additional kind of payment metadata manually handled here? If we need to check metadata for such calls then it should be additional payment type because it is actually is. And if there is no case for `NoOpInterceptor` it should be deleted.

handler/interceptors.go
```diff
@@ -372,9 +375,47 @@ func GetSingleValue(md metadata.MD, key string) (value string, err *GrpcError) {
...
+
+	}
 	return handler(srv, ss)
 }
+func checkIfSignerIsAnAllowedUser(context *GrpcStreamContext) (err error) {
+	signature, grpcErr := GetBytes(context.MD, PaymentChannelSignatureHeader)
+	if grpcErr != nil {
+		return fmt.Errorf("Metadata %v is not set correctly for curation requests",PaymentChannelSignatureHeader)
+	}
+	blockNumber, grpcErr := GetBigInt(context.MD, CurrentBlockNumberHeader)
```
vsbogd: Checking `blockNumber` when blockchain is disabled is impossible so I don't see a reason to include it in header. In any case the first intercepted package can be used for authorization.

handler/interceptors.go
```diff
@@ -375,6 +377,43 @@ func NoOpInterceptor(srv interface{}, ss grpc.ServerStream, info *grpc.StreamSer
...
+		return err.Err()
+	}
+	if err := checkIfSignerIsAnAllowedUser(context); err != nil {
+		return err
+	}
+	return handler(srv, ss)
+}
+
+func checkIfSignerIsAnAllowedUser(context *GrpcStreamContext) (err error) {
+	signature, grpcErr := GetBytes(context.MD, AllowedUserSignatureHeader)
```
vsbogd: I would also add separate `snet-payment-type` value for such type of authorization in order to have consistent metadata format for all cases. Thus `snet-payment-type` can always be used to understand which authorization procedure is applicable to the header and which fields are available.

snetd/cmd/components.go
```diff
@@ -394,6 +405,11 @@ type VerifyMeteringResponse struct {
 
 func (components *Components) GrpcPaymentValidationInterceptor() grpc.StreamServerInterceptor {
 	if !components.Blockchain().Enabled() {
+		if config.GetBool(config.AllowedUserFlag) {
+			log.Info("Blockchain is disabled And AllowedUserFlag is enabled")
+			return handler.AllowedUserValidationInterceptor(components.AllowedUserPaymentHandler())
```
vsbogd: After implementing separate payment handler additional interceptor is not needed anymore:
```suggestion
			return handler.GrpcPaymentValidationInterceptor(components.AllowedUserPaymentHandler())
```
You can delete `allowedUserValidationInterceptor` and all related functions.

Issue #472, opened 2020-03-12
Authenticating Users for Daemon Configuration Services 


anandrgitnirman:
Currently Daemon supports only one authenticating User who can access the configuration service 
ability to see Daemon configuration , Stop Daemon , Pause Daemon , See if any requests are in progress , edit Daemon configurations are the functions supported by configuration service

 , make this as a list of user , this gives the service provider better options to configure who all can make access the configuration services

PR #473, opened 2020-03-12
#472 : Autorized users for access to configuration service 

anandrgitnirman:
#472

### DIFF
--- a/configuration_service/configuration_service.go
+++ b/configuration_service/configuration_service.go
@@ -7,24 +7,42 @@ import (
 	"github.com/ethereum/go-ethereum/accounts/abi"
 	"github.com/ethereum/go-ethereum/common"
 	"github.com/singnet/snet-daemon/authutils"
-	"github.com/singnet/snet-daemon/blockchain"
 	"github.com/singnet/snet-daemon/config"
 	log "github.com/sirupsen/logrus"
 	"golang.org/x/net/context"
 	"math/big"
 	"sort"
-	"strings"
 )
 
 type ConfigurationService struct {
-	//Has the authentication address that will be used to validate any incoming requests for Configuration Service
-	address string
-	broadcast *MessageBroadcaster
+	//Has the authentication authenticationAddressList that will be used to validate any incoming requests for Configuration Service
+	authenticationAddressList []common.Address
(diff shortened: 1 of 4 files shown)


configuration_service/configuration_service.go
```diff
@@ -78,61 +96,51 @@ func (service ConfigurationService) IsDaemonProcessingRequests(ctx context.Conte
...
 
+	signerFromMessage, err := authutils.GetSignerAddressFromMessage(service.getMessageBytes(prefix, auth.CurrentBlock), auth.GetSignature())
+	if err != nil {
+		log.Error(err)
+		return err
+	}
 	//Check if the Signature is Valid and Signed accordingly
-	if err = authutils.VerifySigner(service.getMessageBytes(prefix, auth.CurrentBlock),
-		auth.GetSignature(), blockchain.HexToAddress(service.address)); err != nil {
+	if err = service.checkAuthenticationAddress(*signerFromMessage); err != nil {
```
raamb: Do we not need to verifySigner anymore?

configuration_service/configuration_service.go
```diff
@@ -78,61 +96,51 @@ func (service ConfigurationService) IsDaemonProcessingRequests(ctx context.Conte
...
-	}
-
 	return service
 }
 
-//Message format has been agreed to be as the below ( prefix,block number,and authenticating address)
-func (service *ConfigurationService) getMessageBytes(prefixMessage string, blocknumber uint64) []byte {
+//Message format has been agreed to be as the below ( prefix,block number,and authenticating authenticationAddressList)
+func (service *ConfigurationService) getMessageBytes(prefixMessage string, blockNumber uint64) []byte {
 	message := bytes.Join([][]byte{
```
raamb: service.address is not used in the message?

configuration_service/configuration_service.go
anandsnet: We do, just that it's not one address but a list of addresses that we need to loop and check.hence a different method

configuration_service/configuration_service.go
anandsnet: Not required , you derive the address from the signature ,

Issue #474, opened 2020-03-23
Setting the gRPC message legth of the Daemon proxy.


anandrgitnirman:
We need to parametrize the max send and recv message for the passthrough endpoint point 

The below error was reported when you have large files to send / recv 

Error: <_InactiveRpcError of RPC that terminated with:
	status = StatusCode.RESOURCE_EXHAUSTED
	details = "grpc: received message larger than max (7062886 vs. 4194304)"
	debug_error_string = "{"created":"@1584804270.704719740","description":"Error received from peer ipv4:54.203.198.53:7067","file":"src/core/lib/surface/call.cc","file_line":1056,"grpc_message":"grpc: received message larger than max (7062886 vs. 4194304)","grpc_status":8}"
>
If you want to see full Traceback then run:
snet --print-traceback [parameters]

PR #475, opened 2020-03-23
#474 : We need to parametrize the max send and recv message for the passthrough endpoint point

anandrgitnirman:
We need to parametrize the max send and recv message for the passthrough endpoint point
#474


anandrgitnirman:
https://godoc.org/google.golang.org/grpc#MaxCallRecvMsgSize

Issue #487, opened 2020-04-06
set passthrough_enabled = true by default 


anandrgitnirman:
by default enforce this behavoir 

- set passthrough_enabled = true 

- the user also needs to specify a valid URL for the config passthrough_endpoint

user has the choice to set the value to false explicitly

PR #488, opened 2020-04-06
#487 :set passthrough_enabled = true by default 

anandrgitnirman:
by default enforce this behavoir

set passthrough_enabled = true

the user also needs to specify a valid URL for the config passthrough_endpoint

user has the choice to set the value to false explicitly

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -310,7 +310,7 @@ func ValidateEndpoints(daemonEndpoint string, passthroughEndpoint string) error 
 
 	passthroughURL, err := url.Parse(passthroughEndpoint)
 	if err != nil || passthroughURL.Host == "" {
-		return errors.New("passthrough_endpoint must be a valid URL")
+		return errors.New("passthrough_endpoint is the endpoint of your AI service in the daemon config and needs to be a valid url.")
 	}
 	daemonHost, daemonPort, err := net.SplitHostPort(daemonEndpoint)
 	if err != nil {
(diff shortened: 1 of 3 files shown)


arturgontijo:
Cool!
My idea was is one sets `passthrough_endpoint` with a valid URL, Daemon should automatically sets `passthrough_enabled` to true.
If I got your update right, it's doing it. =)

config/config.go
```diff
@@ -307,9 +307,10 @@ func ValidateEmail(email string) bool {
 }
 
 func ValidateEndpoints(daemonEndpoint string, passthroughEndpoint string) error {
+
 	passthroughURL, err := url.Parse(passthroughEndpoint)
-	if err != nil {
-		return errors.New("passthrough endpoint must be a valid URL")
+	if err != nil || passthroughURL.Host == "" {
```
raamb, line 312: Lets add the isValidURL check as well?

config/config.go
```diff
@@ -307,9 +307,10 @@ func ValidateEmail(email string) bool {
...
+		return errors.New("passthrough_endpoint must be a valid URL")
```
raamb, line 313: Lets make the error message descriptive. Something like- You need to set the passthrough_endpoint variable in the daemon config to the endpoint of your AI service"

config/config.go
anandrgitnirman: passthroughURL, err := url.Parse(passthroughEndpoint) , this line takes care of it  
isValidURL will just return a boolean , 
and we also need to do additiona checks in the function from the passthroughURL struct 
For example 
the port number after parsing to make sure the daemon port and the service port are not the same

config/config.go
anandrgitnirman: done

Issue #490, opened 2020-04-18
Make sure no sensitive config  is displayed in logs are returned back as response 


anandrgitnirman:
give a place holder to mark keys as hidden

PR #491, opened 2020-04-18
#490 :hide any sensitve info from logs / response sent

anandrgitnirman:
hide any sensitve info from logs / response sent

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -274,13 +274,20 @@ func SubWithDefault(config *viper.Viper, key string) *viper.Viper {
 	return sub
 }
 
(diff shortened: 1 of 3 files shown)


config/config.go
```diff
@@ -274,13 +274,20 @@ func SubWithDefault(config *viper.Viper, key string) *viper.Viper {
 	return sub
 }
 
+var HiddenKeys = map[string]bool{
```
raamb, line 277: Can we do it the other way. Maintain the list of display keys and only send that. It's safer

anandrgitnirman: agrre, done

Issue #495, opened 2020-05-08
Simplify signatures for Curation  / (  Allowed User mode when block chain is disabled )


anandrgitnirman:
the allowed user mode was introduced to let the service provider decide if only certain users need to be able to make calls . This feature came in very handy specially during the curation mode , ( i.e the service is not yet published in blockchain , but the certain users will still be able to access and review ).
However this came in with a cost, now the clients need to be aware of certain signature supported just for this curation mode. a client would not know if the Daemon is running on a block chain enabled or disabled mode.This issue is to track the followig 
1) to get away from having a seprate signture but still have the feature 
2) and re use the signature for the paid mode, ( this way clients are agnostic and will sign in the standard way )
, thus getting away from the cost of building a need for a new signature in snet-cli/dapp/SDKs we support

PR #496, opened 2020-05-09
#495 : Simplify / Reduce sigatures ( get away from seprate signature for alloweduser) 

anandrgitnirman:
#495 
Please note , we  need MPE Address to verify the signature , adding this as a header across all clients is a simpler change compared to asking every client to support a  new signature for curation mode ( Block chain mode is disbled and allowed user mode enabled ) , for now if you want a client to be able to call Daemon in curation mode, you need to add this header


vsbogd:
Hi @anandrgitnirman , @raamb , I don't understand the scenario. Issue says about service which is not published on the blockchain during curation mode. Previously I thought that "curation" is when service is published on the blockchain but not published on Dapp. Coud you please describe step by step scenario which is assumed for this functionality?

anandrgitnirman:
> Hi @anandrgitnirman , @raamb , I don't understand the scenario. Issue says about service which is not published on the blockchain during curation mode. Previously I thought that "curation" is when service is published on the blockchain but not published on Dapp. Coud you please describe step by step scenario which is assumed for this functionality?

hi @vsbogd , 
If a service is curated ( rather approved) it becomes visible on Dapp. 
After a serivce is reviewed and approved from the publisher portal , the users can  publish it on block chain from the Publiher portal ( easy steps!), the moment this is done , Dapp backend knows that this service was published from Publisher portal and hence is auto curated.The curation flag is set to zero by default for any serivces published from the snet-cli.We would manually go and curate this so that it comes in dapp ( for the service deployed before the launch of publisher portal) 

An approver needs to test the service before it is published on mainnet from our publisher portal 
we dont want to publish this service on ropsten as it becomes accessible publicly .Hence a daemon is spawend on a block chain disable mode .at the same time you dont want this to be accessd by anyone , hence only certain users ( allowed users) defined in the config can make calls to daemon
Call to daemon happens from our existing clients ( Dapp, SDK and snet-cli) , who should be agnostic to Daemon running in block chain endabled mode or disabled mode and call daemon in a standard way.This will now be acheived with this PR

anandrgitnirman:
Curation mode to us, more or less means ( it is ready for review/being reviewed )

vsbogd:
As discussed in curation mode app is published as part of the special Registry contract instance so it can be used by SDK with contract address which is different from default. This mode is necessary to test application being published from Dapp and snet-cli so supporting this mode generally in SDK is not critical.

On the other hand I like the idea of sending MPE contract address along other payment data. The only reason why is not implemented from the very beginning is a desire to make header as small as possible.

Issue #309, opened 2019-07-08
Enable Concurrent calls in Daemon : State Service Enhancements / New Proto 


anandrgitnirman:
It is expected that the user would call the GetChannelState to Determine the Current state of the Channel
Based on the usage forecast, the user/client will have to sign for an amount L + U , where L is the last amount Signed
and U is the amount based on expected usage.
Please be aware that the Signing up an amount upfront ( Pre Paid)  does come with a risk and hence the
user must exercise caution on the amount signed specially with new service providers.
If there is no need of making concurrent calls then you may consider pay per mode.
Using a Token, the Client can now make concurrent calls, which was not supported previously with the pay per use

PR #497, opened 2020-05-19
Token state

anandrgitnirman:
#309


anandrgitnirman:
lets also look on how the clients will be impacted
old client - new Daemon  ( this should still be supported)

escrow/token_service.proto
```diff
@@ -0,0 +1,61 @@
...
+    bytes signature = 4;
+
+}
+
+// TokenReply message contains a latest channel state. current_nonce and
+message TokenReply {
+    // current_nonce is a latest nonce of the payment channel.
+    uint64 channel_id = 1;
+
+    // current_signed_amount is a last amount which were signed by client with current_nonce
```
vsbogd, line 50: Looks like copy-pasted comment here and line below.

escrow/token_service.proto
```diff
@@ -0,0 +1,61 @@
...
+//However the pay per mode is a lot secure than the pre-paid mode.
+service TokenService {
+    // GetToken method checks the Signature sent and returns a Token
+    // 1) The Signature is valid and has to be signed in the below format
+    //"__MPE_claim_message"+MpeContractAddress+ChannelID+ChannelNonce+SignedAmount
+    //Signature is to let the Service Provider make a claim
+    // 2) Signed amount >= Last amount Signed.
+    //planned amount 10 , usage 5 , this signed 10 cogs
+    //  if Signed amount == Last Signed amount , then check if planned_amount < used_amount
+    //  if Signed amount > Last Signed amount , then update the planned amount = planned amount + Signed Amount
```
vsbogd, line 22: Not sure what is a `plannned amount` here. Am I correct that in situation when user already spent 5 cogs from channel and he wants to sign upfront he signs 13 cogs for example and this means that `planned_amount` is 13 and `used_amount` is 5 ?

escrow/token_service.proto
```diff
@@ -0,0 +1,61 @@
...
+//Using a Token, the Client can now make concurrent calls, which was not supported previously with the pay per mode.
+//However the pay per mode is a lot secure than the pre-paid mode.
+service TokenService {
+    // GetToken method checks the Signature sent and returns a Token
+    // 1) The Signature is valid and has to be signed in the below format
+    //"__MPE_claim_message"+MpeContractAddress+ChannelID+ChannelNonce+SignedAmount
+    //Signature is to let the Service Provider make a claim
+    // 2) Signed amount >= Last amount Signed.
+    //planned amount 10 , usage 5 , this signed 10 cogs
+    //  if Signed amount == Last Signed amount , then check if planned_amount < used_amount
```
vsbogd, line 21: Could you please provide an example? I didn't get it.

escrow/token_service.proto
anandrgitnirman: fixed

escrow/token_service.proto
anandrgitnirman: thanls @vsbogd , just to be sure
amount deposited in channel = 100 cogs 
amount used = 5 Cogs => the user would have given a sig5 ( signature for 5 cogs) 
so the planned amount so far is 5 cogs and the used amount is 5 cogs so far 

Now the user wishes to Sign for 13 Cogs , the planned amount  = Max Amount Signed 
the planned amount now becomes 13 Cogs as you have mentioned and the used amont is still 5 cogs ( thus the user can now make calls worth 13-5 = 8 Cogs) .

The claims in the current block chain supports planned amount and used/actual amount, so this ties nicely on what the latest MPE contract supports.

the planned amount and used amount against a given channel Id will be maintained in a seprate state in etcd , this way user can contiue to use pay per use / support for concurrent calls simultaneously , the new daemon should smoothly run even on old clients

escrow/token_service.proto
anandrgitnirman: Lets use the same example 
amount deposited in channel = 100 cogs
amount used = 5 Cogs => the user would have given a sig5 ( signature for 5 cogs)
so the planned amount so far is 5 cogs and the used amount is 5 cogs so far

Now the user wishes to Sign for 13 Cogs , the planned amount = Max Amount Signed = 13
the planned amount now becomes 13 Cogs as you have mentioned and the used amont is still 5 cogs ( thus the user can now make calls worth 13-5 = 8 Cogs) .

The user happily makes 3 successful calls now => used calls = 8 calls now => balance calls left = 13-8 = 5, now the Token expires .

The user makes a call to State Service to get the latest channel state and sees that the last amount signed was 13 cogs and still has balance of 5 calls , now  the user needs a new Token, does the user again have to sign for a new amount ? no, the user just re sends the last Sig(Sig13) he signed to GetToken and the user will be issued a new token.

For Security reasons , should we add a new Sigature for the Get token , with a block number ? 
So the new Signature will be made of the Claim Signature + block number ( everytime we introduce  a new signature , it has an impact on all clients to be able to support this new change ) , hence not considered it yet .. need your thoughts on this too !

escrow/token_service.proto
vsbogd: Ok, then I believe the following:
> if Signed amount > Last Signed amount , then update the planned amount = planned amount + Signed Amount

should be replaced by:
if Signed amount > Last Signed amount , then update the planned amount = Signed Amount

escrow/token_service.proto
vsbogd: Ok, I understand the example but what does the following statement mean:
> then check if planned_amount < used_amount
 
Did you mean: then check "if planned_amount > used_amount"

vsbogd: > For Security reasons , should we add a new Sigature for the Get token , with a block number ?

What is the token expiration time? It is some Ethereum block or UTC timestamp? As far as I see user cannot set it when requesting token?

escrow/token_service.proto
anandrgitnirman: thanks @vsbogd , corrected the comment

escrow/token_service.proto
anandrgitnirman: User cannot set it in current state  , was planning to set this as a configuration in Daemon and default it to say 5 minutes of expiry ( UTC time)

anandrgitnirman: > heck if planned_amount < used_

removed this comment, it seems to cause confusion , moreover we have clearly indicated 
Signed amount >= Last amount Signed.

vsbogd: Ok, it makes sense to add information about timeout into API comments.

I am not sure about authorization. If token expiration time is needed to prevent token stealing then probably we need authorize user each time when token is requested. Otherwise token request could be stolen instead of token. Personally I don't like authorization using block number I explained reasons in https://github.com/singnet/snet-daemon/issues/498. But for the Ethereum escrow workflow this approach looks ok as both parties deal with Ethereum contracts anyway.

Not sure about this comment:
> everytime we introduce a new signature , it has an impact on all clients to be able to support this new change

because as far as I understand this signature (Claim + block number) will be used only in token requesting workflow (it will not affect old paying workflow). Token requesting workflow is a new workflow and it should be supported by clients anyway. Do you mean clients will not be able to reuse previously written code to sign token requests? I don't see it as a big issue because anyway we have different signatures for different cases.

anandrgitnirman: Thanks @vsbogd , then in that case lets introduce an authorization each time a token is requested at the very least we have Auth Sigature   of ( Claim Sig + Block Number)

Issue #311, opened 2019-07-08
Enable Concurrent calls in Daemon : Modification of existing flows to store Payments / Claims


anandrgitnirman:
#308 
payment flows to support concurrency and also be compatible with simultaneous calls to pay per use too

Issue #310, opened 2019-07-08
Enable Concurrent calls in Daemon : Maintaining / Tracking new State in etcd  


anandrgitnirman:
Support custom CAS operation in a single transaction 
Ensure the update is ALWAYS on the latest state read
Support to keep track of how much was planned, how much was used and how much has to be refunded.

anandrgitnirman:
this ticket had some draft commits , closing this as a duplicte of 
#502

PR #500, opened 2020-05-21
Prepaid - Payment flow 

anandrgitnirman:
#311  
#310 
Once the other PRs are merged , will fill in the todo -list 
Also have marked a few places with //todo , to  see if there are better ways to address this .


anandrgitnirman:
Will raise an other one , with a simplified approach

Issue #312, opened 2019-07-08
Enable Concurrent calls in Daemon : Validations 


anandrgitnirman:
#308 
One of the first step to enable concurrecy is to generate a Token and also validate the token received 
keepnig in mind  , that there can be multiple daemons from the same group , so token issued by first daemon in the group can be used to make a service call to say Daemon 2 in the group

PR #508, opened 2020-08-02
Token Validations + Blockchain Disabled mode for Token Service  (  on Service call) 

anandrgitnirman:
#312


config/config.go
```diff
@@ -120,7 +120,8 @@ const (
 		"enabled": false
 	},
 	"alerts_email": "", 
-	"service_heartbeat_type": "http"
+	"service_heartbeat_type": "http",
+    "token_expiry_in_seconds": 20
```
raamb, line 124: Lets make this in minutes and keep the same default as in the readme file. Also the default value can be 24 hrs

anandrgitnirman: fixed

PR #515, opened 2020-12-02
Pass the Org Id and the Service Id for channel state status 

anandrgitnirman:
#514 
This is to keep track of which service was using the balance from channel for better auditing


vsbogd (approved):

ksridharbabuus (approved):
Approved.

Issue #526, opened 2021-01-17
Support Dynamic Pricing in Daemon 


anandrgitnirman:
Curretly all payment checks in Daemon are based on fixed pricing 
We need to support Dynamic prcing ( i.e compute the price of the service call ) , again this is something best defined by the service provider, however we need to have tool/ support this in Daemon 
For example you call a 
PackageA.ServiceA.Method1 ( RequestParams) returns (OutPutParms)
Now price to call Method1 depends on what RequestParms are ( ex file size, computational complexity etc) , this is again best determined by the service Provider . 

so here is the proposal 
  Use method Level Options to tell the method to call to determine Price 
 PS , this method is expected to have the exact same params as the orginal service call method, but will return a standrad message having price . This price will be used by Daemon to do all the validations 

**Algorithm**
   - Compute the PriceDetermination methods mapped or to be called for service methods when Dynamic pricing is enabled
   - when a service calls comes in 
       - get the pricing method to call
       - use the same i/p to invoke this pricing method
       - get the price to be used 
       - perform validations in Daemon 
       - if all validtions are passed on payment , make the actual service call 

```proto
syntax = "proto3";
import "google/protobuf/descriptor.proto";
package escrow;
extend google.protobuf.MethodOptions {
    EstimatePrice my_method_option = 9999127;
}
//In Order to Support Dynamic Pricing , we need a way to indicate which of the methods will need dynamic pricing Support
//By Dynamic Pricing , the pricing of the service call will be a variable of the inputs passed.
//The Service Provider is best positioned to determine this
//The below standards are to be followed
// 1) The Input Message for the Pricing RPC call will be exactly same as the Input Message of the Service Call
// 2) The Output Message for the Pricing RPC MUST return PriceInCogs.
// 3) Define the name of the RPC method to be called using Method Option EstimatePrice


//helps determine which method to call for dynamic pricing,
//format is of type "packageName/serviceName/MethodName", Example :"/example_service.Calculator/estimate_add"
//Daemon will invoke the actual RPC method , but will invoke the method defined in the method options before
//to determine the price that needs to be charged for this call
message EstimatePrice {
    string estimatePriceMethod = 1;
}
//The Value returned by  the Pricing call
message PriceInCogs {
    uint64 price = 1;
}
```

As an example 
```proto

syntax = "proto3";
import "google/protobuf/descriptor.proto";
import "github.com/singnet/snet-daemon/pricing/pricing.proto";// you need to import both these protos
package example_service;

[...]

PR #527, opened 2021-01-18
Support for Dynamic Pricing in Daemon 

anandrgitnirman:
PS see #526 for details


handler/grpc.go
```diff
@@ -252,6 +266,59 @@ func (g grpcHandler) grpcToJSONRPC(srv interface{}, inStream grpc.ServerStream)
...
+	sentMessage      interface{}
+}
+
+func (f *wrapperServerStream) SetTrailer(md metadata.MD) {
+	f.stream.SetTrailer(md)
+}
+
+func NewWrapperServerStream(stream grpc.ServerStream) (grpc.ServerStream, error) {
+	m := &codec.GrpcFrame{}
+	err := stream.RecvMsg(m)
```
vsbogd, line 282: Why receiving message is needed here?

handler/interceptors.go
```diff
@@ -235,10 +244,72 @@ func GrpcPaymentValidationInterceptor(defaultPaymentHandler PaymentHandler, paym
 }
 
 type paymentValidationInterceptor struct {
+	serviceMetadata       *blockchain.ServiceMetadata
 	defaultPaymentHandler PaymentHandler
 	paymentHandlers       map[string]PaymentHandler
 }
 
+func (interceptor *paymentValidationInterceptor) checkForDynamicPricing(srv interface{}, ss grpc.ServerStream,
```
vsbogd, line 252: My original intention was isolating pricing related logic in `pricing` module. In particular such implementation will allow get price dynamically only when it is needed (i.e. don't do this when `free-call` or `authorized-user` payment strategy is applied).

If I would implement this I would do the following:
- move checking if dynamic pricing is applicable to `PricingStrategy.determincePricingApplicable` method
- add new `DYNAMIC_PRICING` into `PriceType`
- add new `dynamic_method_pricing.go` and move getting price related code into it (similar to `fixed_method_pricing.go`
- to make above possible probably we need to modify `handler.GrpcStreamContext` and add additional information into it if needed
- `ServiceMetadata` should be part of the `PricingStrategy` I believe, but I am not sure how to implement it better

handler/grpc.go
anandrgitnirman: the same message will be used to make a call to estimate price and also to the actual service invocation

handler/interceptors.go
anandrgitnirman: made changes , if it looks fine , will check in the test cases as well

handler/grpc.go
```diff
@@ -174,7 +175,20 @@ func forwardServerToClient(src grpc.ServerStream, dst grpc.ClientStream) chan er
...
+			//future calls will result in io.EOF , we want to retrieve the
+			//the first message sent by the client and pass this on the regualar service call
+			//This is done to be able to make calls to support regualr Service call + Dynamic pricing call
+			if i == 0 {
+				//todo we need to think through to determine price for every call on stream calls
+				//will be handled when we support streaming and pricing across all clients in snet-platform
+				if wrappedStream, ok := src.(*wrapperServerStream); ok {
+					f = (wrappedStream.OriginalRecvMsg()).(*codec.GrpcFrame)
+				} else {
+					src.RecvMsg(f)
```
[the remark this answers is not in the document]
anandrgitnirman: fixed this

vsbogd (approved):

vsbogd:
Looks good to me, thanks

Issue #504, opened 2020-07-01
clean up storage dependency 


anandrgitnirman:
etcddb should not depend on escrow. There is AtomicStorage interface which is declared in atomic_storage.go but we missed it before because Go doesn't require importing interface before implementing it. I would suggest making new storage package, moving atomic_storage.go, memory_storage.go and etcddb under this package. Then both etcddb and escrow will depend on `storage.

PR #530, opened 2021-01-22
Storage - move AtomicStorage to a new package 

anandrgitnirman:
#504


ksridharbabuus (commented):
Overall looks fine.
Sometime in review itself I am not able to figure it out between storage2 vs storage reference.
Is it possible to provide appropriate name.

etcddb/etcddb_test.go
```diff
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"github.com/singnet/snet-daemon/blockchain"
 	"github.com/singnet/snet-daemon/escrow"
```
vsbogd, line 7: Looks like `escrow` dependency here is kept only because of `func (suite *EtcdTestSuite) TestExecuteTransaction() {
` test. Is it possible to move these test under the `escrow` package?

anandrgitnirman: fixed

ksridharbabuus (approved):
Approved.

etcddb/etcddb_test.go
vsbogd: @anandrgitnirman , I am very sorry if I was unclear, but I meant only `TestExecuteTransaction` test. It looked like it is only test which depends on `escrow` package and it can be moved into `prepaid_service_test.go` file. But I looked carefully on the test and I see it actually tests `EtcdClient.ExecuteTransaction` method and uses `PrepaidStorage` only as a class which uses transactions. So I think it is better to keep tests in `etcddb` and proper way is to add two different tests: one for `EtcdClient.ExecuteTransaction` and another for `PrepaidStorage`. Or make `TestExecuteTransaction` kind of integration test (which we don't have yet).

Would you mind removing last commit I think for now it is more logical.

anandrgitnirman: @vsbogd , makes sense , addressed this

anandrgitnirman: #531

vsbogd (approved):

Issue #429, opened 2019-12-02
Alternative to Mockables to run tests without net 


anandrgitnirman:
While mockbales enabled testing the requet / response it did come back with an issue on always internet to run yuur test cases .
This ticket is to look for alternatives and eventually be able to run all unit tests locally ( even without internet)

PR #566, opened 2022-04-13
#426 - get away from mockables

anandrgitnirman:
Fix the build and get away from mockables 
#429

### DIFF
--- a/metrics/clients_test.go
+++ b/metrics/clients_test.go
@@ -90,32 +90,25 @@ func (suite *ClientTestSuite) Test_callgRPCServiceHeartbeat() {
 }
 
 func (suite *ClientTestSuite) Test_callHTTPServiceHeartbeat() {
-	serviceURL := "http://localhost:1111/heartbeat"
+	serviceURL := suite.serviceURL + "/heartbeat"
 	heartbeat, err := callHTTPServiceHeartbeat(serviceURL)
 	assert.False(suite.T(), err != nil)
 	assert.NotEqual(suite.T(), string(heartbeat), `{}`, "Service Heartbeat must not be empty.")
 	assert.Equal(suite.T(), string(heartbeat), `{"serviceID":"SERVICE001","status":"SERVING"}`,
 		"Unexpected service heartbeat")
 
-	/*	var sHeartbeat pb.HeartbeatMsg
-		err = json.Unmarshal(heartbeat, &sHeartbeat)
-		assert.True(t, err != nil)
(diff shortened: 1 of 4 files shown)


training/training.proto
```diff
@@ -0,0 +1,60 @@
...
+
+message ModelId {
+  string model_id = 1;
+  string method_name = 2;
+}
+
+enum Status {
+  CREATED = 0;
+  IN_PROGRESS = 1;
+  UN_KNOW_ERROR =2;
```
raamb, line 12: Lets just call it ERROR

training/training.proto
```diff
@@ -0,0 +1,60 @@
...
+}
+
+service Model {
+
+  // The AI developer needs to Implement this service and Daemon will call these
+  rpc create_model(CreateModelRequest) returns (ModelDetailsResponse) {}
+  rpc update_model_access(UpdateModelRequest) returns (ModelDetailsResponse) {}
+  rpc delete_model(UpdateModelRequest) returns (ModelDetailsResponse) {}
+  rpc get_model_details(ModelId) returns (ModelDetailsResponse) {}
+  rpc training_status (ModelId) returns (ModelDetailsResponse) {}
```
raamb, line 58: Can we rename this to "get_training_status"

metrics/clients_test.go
```diff
@@ -53,49 +73,49 @@ func StartMockGrpcService() {
...
 	serviceURL = "localhost:26000"
 	heartbeat, err = callgRPCServiceHeartbeat(serviceURL)
-	assert.True(t, err != nil)
+	assert.True(suite.T(), err != nil)
 }
 
-func Test_callHTTPServiceHeartbeat(t *testing.T) {
-	serviceURL := "http://demo5343751.mockable.io/heartbeat"
+func (suite *ClientTestSuite) Test_callHTTPServiceHeartbeat() {
+	serviceURL := "http://localhost:1111/heartbeat"
```
raamb, line 93: Lets make the port dynamic - via config file etc

anandrgitnirman: have removed the hardcoded endpoint from multiple places, the service url end point is specified as a config file , for a test case , dont wanna add a new config for the port number

anandrgitnirman: [e949ae6](https://github.com/singnet/snet-daemon/pull/566/commits/e949ae6e5b0789b26553a46c7d341f25a6cb082e) - latest commit

raamb (approved):

Issue #565, opened 2022-04-13
Support for Training Models  in Daemon 


anandrgitnirman:
We are looking at enhancing the Daemon to Support /standardize 

- Creation of Training models
- Ensure only certain address can call training methods
- Check the status of the model being trained

semyon-dev:
I'm closing it due to the fact that support for the training functionality has been implemented

PR #567, opened 2022-05-11
#565

anandrgitnirman:
Grpc server impl of training.proto - Skeleton code checkin


training/service.go
```diff
@@ -0,0 +1,366 @@
...
+	key := service.getModelKeyToUpdate(request.ModelDetailsRequest)
+	data, ok, err := service.storage.Get(key)
+	if ok && err != nil {
+		data.Status = "DELETED"
+		err = service.storage.Put(key, data)
+	}
+	return
+}
+
+func (service ModelService) getModelDetails(request *UpdateModelRequest, response *ModelDetailsResponse) (data *ModelUserData, err error) {
```
raamb, line 97: Remove blank lines to keep it consistent with other methods

training/service.go
```diff
@@ -0,0 +1,366 @@
...
+		log.WithError(fmt.Errorf("issue with retrieving model data from storage"))
+	}
+	return
+}
+func (service ModelService) GetAllModels(c context.Context, request *AccessibleModelsRequest) (response *AccessibleModelsResponse, err error) {
+	if request == nil || request.Authorization == nil {
+		return &AccessibleModelsResponse{Status: Status_ERROR},
+			fmt.Errorf(" INVALID  REQUEST DETAILS , no Authorization provided ")
+	}
+	if err = service.verifySignerForCreateModel(request.Authorization); err != nil {
```
raamb, line 176: Why are we calling "verifySignerForCreateModel" - we are not creating a model here correct? if "verifySignerForCreateModel" is the generic method that checks if caller has access to the model, lets give it a generic name such as checkModelAccess or anything else better

training/service.go
```diff
@@ -0,0 +1,366 @@
...
+		return &AccessibleModelsResponse{Status: Status_ERROR},
+			fmt.Errorf(" authentication FAILED , %v", err)
```
raamb, line 178: Lets change the error message to "Unable to access model"

training/service.go
```diff
@@ -0,0 +1,366 @@
...
+			return response, fmt.Errorf("issue with storing Model Id in the Daemon Storage %v", err)
+		}
+		deferConnection(conn)
+	} else {
+		return &ModelDetailsResponse{Status: Status_ERROR}, fmt.Errorf("error in invoking service for Model Training")
+	}
+	return
+}
+
+//message used to sign is of the form ("__create_model", mpe_address, current_block_number)
```
raamb, line 308: Lets not have different methods, instead have the caller pass in the prefix which should be a constant

training/service.go
anandrgitnirman: signture comprises of 
methodname, signer_address,blocknumber 
Any method called needs to be signed ,verify*methods are for signture verification , will rename it to verifySignature..rather than verifySigner

training/service.go
anandrgitnirman: makes sense will do

raamb (approved):

PR #570, opened 2022-06-02
#565

anandrgitnirman:
To support method level options to identify which serivce methods are training methods and direct these calls to a separate End point


ksridharbabuus (approved):
Approved.

Issue #572, opened 2022-06-20
Add the client storage cert details as part of heartbeat 


anandrgitnirman:
enhance the hearbeat by adding the content shown below in bold
**`"storageClientCertDetails": {
"validFrom": "Valid Since: 2022-05-20 11:55:00 +0000 UTC days",
"validTill": "Valid Till: 2023-05-20 11:55:00 +0000 UTC days"
}`**

```
{
"daemonID": "9afe2405424bac5653f0645f2d9baa3618d9dc916fa50eab926f7f43d10a89d3",
"timestamp": "1655722159",
"status": "Online",
"serviceheartbeat": "{\"serviceID\":\"coreference-resolution-service\",\"status\":\"SERVING\"}",
"daemonVersion": "",
"storageClientCertDetails": {
"validFrom": "Valid Since: 2022-05-20 11:55:00 +0000 UTC days",
"validTill": "Valid Till: 2023-05-20 11:55:00 +0000 UTC days"
}
}
```

PR #573, opened 2022-06-20
#572

anandrgitnirman:
add the client storage cert details to the heartbeat


etcddb/etcddb_client.go
```diff
@@ -60,7 +61,7 @@ func NewEtcdClientFromVip(vip *viper.Viper, metaData *blockchain.OrganizationMet
 
 	var etcdv3 *clientv3.Client
 
-	if checkIfHttps(metaData.GetPaymentStorageEndPoints()) {
+	if utils.CheckIfHttps(metaData.GetPaymentStorageEndPoints()) {
```
raamb, line 64: We should remove checkIfHttps from this file

raamb (approved):

PR #680, opened 2026-05-05
Issue 679

### LINKED ISSUE #679: Harden config validation: require non-empty `token_secret_key`
## Description

The `token_secret_key` config field is used by `customJWTokenServiceImpl.CreateToken` and `VerifyToken` as the HMAC-SHA256 key for prepaid-call JWTs. There's no default for it in `defaultConfigJson`, `config.Validate()` doesn't check it, the README says it's optional. If an operator forgets to set it, the daemon starts with no warning and uses an empty HMAC key.

I think this should be treated like the other required config values (SSL paths, endpoints, free-call private key) and rejected at startup when blockchain mode is on. Failing fast is better than silently running with weakened token security.

## Code reference

- `token/jwt.go:36` and `token/jwt.go:45` — HMAC key sourced from `config.GetString(config.TokenSecretKey)`
- `config/config.go` — `defaultConfigJson` doesn't set `token_secret_key`, and `Validate()` doesn't check it
- `README.md` — field documented as optional

## Steps to reproduce

1. Start `snetd` with a config that omits `token_secret_key`.
2. Daemon starts with no warning.
3. The HMAC key used in `customJWTokenServiceImpl` is `[]byte("")`.

Expected: startup should fail when the field is unset and `blockchain_enabled` is true.

Niek-Kamer:
## Description

Adds startup validation for `token_secret_key` so misconfigured deployments fail fast instead of running with an empty HMAC key for prepaid-call JWTs.

Closes #679

## Changes

`config/config.go` — in `Validate()`, return an error when `token_secret_key` is empty or shorter than 32 bytes and `blockchain_enabled` is true. Same shape as the existing checks for SSL paths and the free-call private key.

## Interface changes

Operators running with `blockchain_enabled: true` and no `token_secret_key` will now see the daemon refuse to start with a clear error message. Previously it started silently.


semyon-dev (approved):

semyon-dev (approved):

Niek-Kamer:
The failing sonar / Build and Analyze check isn't related to this PR's changes. The job exits in 8s with no steps run, looks like the reusable workflow at singnet/security-workflows/.github/workflows/sonar.yml@master declares its secrets (SONAR_TOKEN, OPENSEARCH_*, etc.) as required, and GitHub doesn't pass repo secrets to fork PRs, so secrets: inherit resolves to empty and the called workflow is rejected at startup. Sonar runs fine on master/dev pushes. Could you re-run via workflow_dispatch on this branch, or push the branch to the upstream repo so it can run with secrets?

semyon-dev (approved):

PR #682, opened 2026-05-11
refactor: use reflect.TypeFor

box4wangjing:
Try to use better api reflect.TypeFor instead of reflect.TypeOf when we have known the type.
More info https://github.com/golang/go/issues/60088

Inspired by https://github.com/singnet/snet-daemon/pull/653  and replace all.


box4wangjing:
@semyon-dev Hi, Could you please review this PR at your convenience? Thank you very much.

semyon-dev (approved):
The migration itself looks safe and preserves the existing value-type semantics (T instead of *T).

One thing worth double-checking is that TypedAtomicStorageImpl intentionally expects non-pointer reflected types, while runtime usage operates on pointers (*ModelKey, *ModelData, etc.). This behavior already existed before the refactor, so the PR does not introduce it, but it's an important invariant for future changes.

semyon-dev:
One small note: the PR description says "replace all", but there are still several handwritten storage-related usages of `reflect.TypeOf(T{})` remaining (for example in `payment_channel_storage.go`, `prepaid_storage.go`, and `license_storage.go`).

The remaining `.pb.go` generated files are probably fine to leave untouched, but it may be worth either:

* updating the PR description to clarify the scope, or
* including the remaining storage-layer replacements for consistency.

kiruxaspb (changes requested):
Reject notes:

The PR says "replace all", but there are still a few storage-related reflect.TypeOf(T{}) usages left (e.g. in payment_channel_storage.go, prepaid_storage.go, and license_storage.go). Either update the description or include those replacements for consistency.

PR #700, opened 2026-09-11
fix(metrics): unconfigured heartbeat reports Warning not Online

### LINKED ISSUE #694: Heartbeat reports Online/SERVING for a service that accepts TCP but cannot serve (unconfigured heartbeat path)
### Summary

When no heartbeat endpoint is configured, `GetHeartbeat` falls back to `tcpPingService`, which
opens a TCP connection and closes it. A service process that is alive and holding its port open,
but can no longer serve requests, satisfies that ping — so the daemon reports `Status: "Online"`
and gRPC `SERVING`.

Accepting a connection is something a dead service does. This is the failure mode where a service
dies but its socket stays open, which is common with a wedged worker, an exhausted pool, a crashed
inference backend behind a live HTTP server, or a container whose main thread is stuck.

### Reproduction

Against `cbfd673`, using the repo's own `GetHeartbeat`, with a local listener that calls `Accept()`
and then does nothing with the connection:

```
heartbeatType=""     -> Status="Online"  err=<nil>
heartbeatType="none" -> Status="Online"  err=<nil>
heartbeatType="tcp"  -> Status="Online"  err=<nil>
```

Baseline arm, a listener that actually serves HTTP 200: `Status="Online"`, as expected. So the
check returns the same answer whether the service works or not.

Test used (dropped into `metrics/`, nothing in the repo modified):

```go
func acceptOnly(t *testing.T) string {
      ln, _ := net.Listen("tcp", "127.0.0.1:0")
      go func() { for { c, err := ln.Accept(); if err != nil { return }; _ = c } }()
      t.Cleanup(func() { ln.Close() })
      return "http://" + ln.Addr().String()
}

func TestHeartbeatUnconfigured(t *testing.T) {
      dead := acceptOnly(t)
      for _, hb := range []string{"", "none", "tcp"} {
              h, err := GetHeartbeat(dead, "", hb, "svc", noTraining, map[string]string{}, noBlock)
              t.Logf("heartbeatType=%-6q -> Status=%q err=%v", hb, h.Status, err)
      }
}
```

(Building the `metrics` package needs `go generate ./blockchain/` and `go generate ./training/`
first; `training`'s generated files are not checked in.)

### Where it comes from

`ValidateHeartbeatConfig` treats an unset type or endpoint as valid and says so:

```go
if heartbeatType == "" || heartbeatType == "none" || heartbeatEndpoint == "" {
    zap.L().Info("heartbeat service endpoint not configured, will be using service endpoint to ping the service")
    return nil
}
```

`GetHeartbeat` then routes `""`, `"none"` and `"tcp"` to `tcpPingService`, and:

```go
if err == nil {
    curResp.Status = "SERVING"
    heartbeat.Status = Online.String()
}
```

`tcpPingService` is `net.DialTimeout` → `Close()` → `nil`.

### Why we think it is worth fixing rather than documenting

The narrow issue is not that a TCP ping is weak — it is that **`"none"`, a value that means "no
heartbeat is configured", produces a positive health signal that is indistinguishable from a
verified one.** A consumer reading `Online` cannot tell "the service answered a health check" from
"nobody configured a health check, so we opened a socket."

### Suggested directions (your call entirely)

1. Report a distinct status for the unconfigured case — `Unknown`, or `Warning`, which already
   exists in the `Status` enum — rather than `Online`.
2. Or have `"none"` mean none: no positive claim, no ping.
3. Or keep the ping but label it in the payload, so consumers can tell the two apart.

The `grpc`, `http` and `https` paths were not exercised and are likely fine; this is scoped to the
fallback.

### What we did not establish

- Whether any deployed daemon actually runs with the heartbeat unconfigured.
- Whether anything downstream acts on `Online` in a way that matters.
- Anything about the rest of the daemon. We read and ran one path.

---

*Method note, deliberately last:* we found this by asking, for each assurance check, "what is the
condition under which this check **must** fail?" and then causing that condition. We keep the
write-up of that method at https://github.com/open-asi-governance/open-asi-governance-forum — it is
a practice note, not a standard, it has never been implemented by anyone outside the project that
wrote it, and its own authors have published and corrected several false claims about it. If you
think the method is just fault injection under a new name, you are close to right and we would
genuinely like to hear it. None of that affects the bug above, which stands on its own reproduction.

MrBeldum:
## Description
Unconfigured heartbeat types (`""` / `"none"`) no longer claim `Online` from a TCP accept. They report `Warning` with `NOT_SERVING` so consumers can tell an unchecked service from a verified one. Explicit `"tcp"` still uses `tcpPingService`.

Fixes #694.

## Changes
- `GetHeartbeat`: route `"none"`/`""` away from `tcpPingService`; keep `"tcp"` on the dial path
- Tests for accept-only listeners (unconfigured → Warning, tcp → Online) plus HTTP path still Online
- Update `TestHeartbeatHandler` expectation to match the unconfigured Warning status

## Interface changes
Daemon heartbeat JSON status for unconfigured types is now `"Warning"` instead of a false `"Online"`. gRPC health `Check` accordingly returns non-SERVING for those configs (same as any non-Online status today).

## Test plan
- [x] `go test ./metrics/ -count=1`

### DIFF
--- a/metrics/heartbeat.go
+++ b/metrics/heartbeat.go
@@ -162,17 +162,22 @@ func GetHeartbeat(serviceEndpoint string, serviceHeartbeatURL string, heartbeatT
 		serviceHeartbeatBytes, err = callHTTPServiceHeartbeat(serviceHeartbeatURL)
 		heartbeat.ServiceHeartbeat = string(serviceHeartbeatBytes)
 		zap.L().Debug("Get heartbeat", zap.String("serviceHeartbeatURL", serviceHeartbeatURL), zap.String("serviceHeartbeatBytes", string(serviceHeartbeatBytes)), zap.Error(err))
-	case "none":
-		fallthrough
 	case "tcp":
-		fallthrough
-	case "":
-		// trying to ping the service with serviceEndpoint
+		// Explicit TCP heartbeat: dial the service endpoint.
 		err = tcpPingService(serviceEndpoint)
 		zap.L().Debug("Get heartbeat [tcpPingService]", zap.String("serviceEndpoint", serviceEndpoint), zap.Error(err))
+	case "none", "":
+		// No heartbeat configured. Do not claim Online from a TCP accept —
+		// a wedged process can still accept connections. Report Warning so
+		// consumers can tell "verified" from "unchecked".
+		heartbeat.Status = Warning.String()
+		zap.L().Debug("heartbeat not configured; not claiming Online via TCP accept",
+			zap.String("heartbeatType", heartbeatType),
+			zap.String("serviceEndpoint", serviceEndpoint))
 	}
 
-	if err == nil {
+	// "none"/"" already set Warning above; only promote Online when a real check succeeded.
+	if err == nil && heartbeatType != "none" && heartbeatType != "" {
 		curResp.Status = "SERVING"
 		heartbeat.Status = Online.String()
 	}
--- a/metrics/heartbeat_test.go
+++ b/metrics/heartbeat_test.go
@@ -136,7 +136,7 @@ func (suite *HeartBeatTestSuite) TestHeartbeatHandler() {
 	assert.True(suite.T(), dHeartbeat.TrainingMetadataData.TrainingInProto)
 	assert.NotNil(suite.T(), dHeartbeat, "heartbeat must not be nil")
 
-	assert.Equal(suite.T(), Offline.String(), dHeartbeat.Status, "Invalid State")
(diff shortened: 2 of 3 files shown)


MrBeldum:
@StephenLReed @semyon-dev This addresses #694: unconfigured heartbeat types (`""` / `"none"`) now report `Warning`/`NOT_SERVING` instead of a false `Online` from a TCP accept.

Fork CI (`build`/`test`/`Sonar`) is `action_required` and needs a maintainer approval to run. Happy to adjust if the Warning vs Online contract should look different.

